PR #2, opened 2017-08-04
Fixes and augments the tests, and breaks out the project structure

dlcarter:
Tests are fleshed out and working, and broke out the main file into its smaller constituents.


lib/painted_rabbit/serializer.rb
```diff
@@ -0,0 +1,16 @@
+class PaintedRabbit::Serializer
+  def initialize(block)
+    @block = block
+  end
+  def self.serialize(&block)
+    @_blah = self.new(block)
+  end
+
+  def self.bleh
```
AllPurposeName: NB: Can we replace this metasyntactic variable with a name representing what's actually happening?

dlcarter: This is not new code - just refactored. And I would love a better variable name, but that's not the scope of this PR. (Just structure change). However, that variable name has been bugging me, but naming it is hard, as I assume that's why @HParker named it as such and it's certainly why I haven't given the thought cycles yet to changing it. Would love suggestions!

PR #5, opened 2017-09-11
ActiveRecord::Relation detection

philipqnguyen:
This code never will resolve to true:
```
unless object.is_a?(ActiveRecord::Base) && object.respond_to?(:klass)
  return object
end
```

An `ActiveRecord::Base` object does not have a `#klass` method. The
`#klass` method is found only in `ActiveRecord::Relation`.

I'm assuming that based on the following code, that what we are trying
to detect is an `ActiveRecord::Relation` object. So I added this
instead `object.is_a?(ActiveRecord::Relation)`.

### DIFF
--- a/lib/painted_rabbit/base.rb
+++ b/lib/painted_rabbit/base.rb
@@ -83,9 +83,7 @@ module PaintedRabbit
     private_class_method :object_to_hash
 
     def self.select_columns(object, view:)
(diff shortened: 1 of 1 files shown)


philipqnguyen:
After some discussion, there needs to be some more thoughts into how we should approach performance optimizing for ActiveRecord objects. There is now an issue here #8.

Closing this PR for now.

PR #6, opened 2017-09-12
Implement an improved View interface

dlcarter:
This PR changes the format of defining views from:
```ruby
identifier :id
field :name
include_in [:normal, :extended] do
  field :description
end
include_in :extended do
  field :color
end
```
to
```ruby
identifier :id
field :name
view :normal do
  field :description
end
view :extended do
  incldue_view :normal
  field :color
end
```

And also implements the ability to exclude fields from views using the interface `exclude :field_name`

### DIFF
--- a/lib/painted_rabbit/base.rb
+++ b/lib/painted_rabbit/base.rb
@@ -57,25 +57,27 @@ module PaintedRabbit
     end
 
     def self.associations(view = :default)
-      views.render_fields(view).select { |f| f.options[:association] }
+      views.fields_for(view).select { |f| f.options[:association] }
     end
 
     def self.include_view(view)
+      current_view.included_views << view
     end
 
     def self.exclude(field_name)
+      current_view.excluded_fields << field_name
     end
 
     def self.view(view_name)
-      current_view = views[view_name]
+      @current_view = views[view_name]
       yield
-      current_view = views[:default]
+      @current_view = views[:default]
     end
 
     private
 
     def self.object_to_hash(object, view:)
-      views.render_fields(view).each_with_object({}) do |field, hash|
+      views.fields_for(view).each_with_object({}) do |field, hash|
         hash[field.name] = field.serializer.serialize(field.method, object, field.options)
       end
     end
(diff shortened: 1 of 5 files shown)


lib/painted_rabbit/base.rb
```diff
@@ -40,7 +36,7 @@ def self.render(object, view: :default)
     # This is the magic method that converts complex objects into a simple hash
     # ready for JSON conversion
     def self.prepare(object, view:)
-      unless views.keys.include? view
+      unless views.has_view? view
```
philipqnguyen: Right here, `views` is actually a `ViewCollection`. Also `view` is not a `View` object, but a view name.
Can we rename the variables here and throughout other places in this file?
So that we have `view_collection` for `ViewCollection` and `view_name` for view names.

lib/painted_rabbit/base.rb
```diff
@@ -142,13 +141,13 @@ def self.jsonify(blob)
...
-      @current_views ||= [:default]
+    def self.current_view
+      @current_view ||= views[:default]
     end
-    private_class_method :current_views
+    private_class_method :current_view
 
     def self.views
-      @views ||= { identifier: {}, default: {} }
+      @views ||= ViewCollection.new
```
philipqnguyen: Should we change this method to `self.view_collection`?

lib/painted_rabbit/base.rb
dlcarter: Good call! I started doing this (a tiny bit) changing instances of `view` to `view_name` but did not go far enough to change it everywhere. I think this will really help clear things up

lib/painted_rabbit/base.rb
dlcarter: 👍

lib/painted_rabbit/view_collection.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    def sortable_fields(view_name)
+      fields = views[:default].fields
+      fields += views[view_name].fields
+      views[view_name].included_views.each do |included_view_name|
+        if view_name != included_view_name
+          fields += sortable_fields(included_view_name)
+        end
+      end
+      fields.delete_if do |f|
```
AllPurposeName: It might be more performant here to iterate over the excluded fields and delete from `fields` as necessary.

lib/painted_rabbit/base.rb
```diff
@@ -2,35 +2,31 @@
 require_relative 'painted_rabbit_error'
```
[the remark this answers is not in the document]
dlcarter: Great idea

PR #10, opened 2017-09-18
Gemspec fix

philipqnguyen:
JSON should be a runtime dependency and not just a development
    dependency.

    We are using required keyword arguments in a number of places
    which is a ruby 2.1+ feature.
 
    Additionally, we are using ruby 2.2.2 and above libraries in development
    so to stay safe, we will require ruby 2.2.2 in the gemspec.


philipqnguyen:
Tagging #13

Issue #7, opened 2017-09-12
Benchmarking


mcclayton:
Since performance is a high priority for our serializer - the sooner we implement some basic benchmarking the better.
https://ruby-doc.org/stdlib-1.9.3/libdoc/benchmark/rdoc/Benchmark.html
We should introduce benchmark tests so that we can see how changes we make effect our performance from PR to PR.

philipqnguyen:
closing this because #12 was merged.

PR #12, opened 2017-09-19
Benchmark tests #7

philipqnguyen:
There are two types of tests; an iterations per second (IPS) test, and a test to see if increasing number of objects passed to `Base#self.render` increases execution time linearly.

We consider both `ActiveRecord` objects and simple objects (`OpenStruct`) when making the tests.

In the IPS tests, we assume that `ActiveRecord` objects will iterate at least 2500 times per second, and an `OpenStruct` object will iterate at least 3000 times per second. Otherwise, the test will fail.

Benchmark tests also only run if you type `bundle exec rake benchmarks`. Default rake will not run the benchmarks.

To see results, open up circle ci, and view the rake benchmarks logs.

### DIFF
--- a/test/benchmarks/benchmark_helper.rb
+++ b/test/benchmarks/benchmark_helper.rb
@@ -1,0 +1,11 @@
+module BenchmarkHelper
+  def iterate(&block)
+    start = Time.now
+    count = 0
+    while Time.now - start <= 1.second do
+      block.call
+      count += 1
+    end
+    count
+  end
+end
--- a/Rakefile
+++ b/Rakefile
@@ -25,9 +25,16 @@ require 'rake/testtask'
 
 Rake::TestTask.new(:test) do |t|
   t.libs << 'test'
-  t.pattern = 'test/**/*_test.rb'
+  t.test_files = Dir['test/**/*_test.rb'].reject do |path|
+    path.include?('benchmarks')
+  end
   t.verbose = false
 end
 
+Rake::TestTask.new(:benchmarks) do |t|
+  t.libs << 'test'
+  t.pattern = 'test/benchmarks/**/*_test.rb'
+  t.verbose = false
+end
 
 task default: :test
--- a/circle.yml
+++ b/circle.yml
@@ -9,3 +9,4 @@ database:
 test:
   post:
     - bundle exec rake
+    - bundle exec rake benchmarks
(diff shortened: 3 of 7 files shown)


philipqnguyen:
Just tagging issue #7 to this PR.

test/benchmarks/benchmark_helper.rb
```diff
@@ -0,0 +1,11 @@
+module BenchmarkHelper
```
mcclayton: Any particular reason we don't want to use the standard benchmark library? http://ruby-doc.org/stdlib-2.3.0/libdoc/benchmark/rdoc/Benchmark.html

philipqnguyen: I was thinking in terms of iterations per second, which isn't part of the standard ruby library. I'm up to discuss this some more and swapping out measuring IPS for the standard benchmark if need be.

mcclayton: ^ Disregard.

mcclayton: Ah, no worries. I think this is good! Thanks again for throwing this together so quickly.

philipqnguyen:
We can't do anything with `Minitest::Benchmark` making ugly outputs in the big o tests so we are leaving it as it is. 

However I have amended d1fcd7a to prepended a newline to the `puts` for the IPS tests here  https://github.com/procore/painted-rabbit/pull/12/commits/d1fcd7a7def31f4eda59e4c8492c004cbea1637b#diff-d9ed00c8599d0f8841057ec89f27887fR18 and here https://github.com/procore/painted-rabbit/pull/12/commits/d1fcd7a7def31f4eda59e4c8492c004cbea1637b#diff-56afaba6067307d4de99c5f8113c8848R18

test/benchmarks/benchmark_helper.rb
```diff
@@ -0,0 +1,11 @@
...
+  def iterate(&block)
+    start = Time.now
+    count = 0
+    while Time.now - start <= 1.second do
+      block.call
```
AllPurposeName: As I learned just now: We should use `yield` instead of `block.call` since `block.call` is almost 2x slower

source: https://stackoverflow.com/questions/1410160/ruby-proccall-vs-yield

philipqnguyen: nice catch. I added an issue for it here https://github.com/procore/painted-rabbit/issues/15

PR #18, opened 2017-09-25
Add Optimizer to query ActiveRecord::Relation

### LINKED ISSUE #8: Optimization for ActiveRecord is not working.
These two singleton methods found in `Base` are not being utilized:

https://github.com/procore/painted-rabbit/blob/9c10456865f607c7c6e0e553bd965b790d63ab5e/lib/painted_rabbit/base.rb#L85-L110

The culprit is that the guard clause `object.is_a?(ActiveRecord::Base) && object.respond_to?(:klass)` will never resolve to true. The reasoning is that `#klass` is a method found in `ActiveRecord::Relation` and not `ActiveRecord::Base`. While this problem does not break anything, the original intention to increase performance for both `ActiveRecord::Relation` and `ActiveRecord::Base` is currently not being achieved.

A few things to consider:
How should we consider optimizing for both `ActiveRecord::Relation` and `ActiveRecord::Base` objects. Do we even need to optimize for both?

philipqnguyen:
Add `Optimizer` to query `ActiveRecord::Relation`

1. detect for `ActiveRecord::Relation` instead of `ActiveRecord::Base`
objects for the time being since all the optimization methods we are
doing are actually `ActiveRecord::Relation` methods and not
`ActiveRecord::Base` methods.

2. Move all optimization related methods to it's own class, `Optimizer`

Resolves #8

### DIFF
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -4,6 +4,7 @@ require_relative 'field'
 require_relative 'serializer'
 require_relative 'view'
 require_relative 'view_collection'
+require_relative 'optimizer'
 require_relative 'serializers/association_serializer'
 require_relative 'serializers/public_send_serializer'
 
@@ -42,7 +43,8 @@ module Blueprinter
       unless view_collection.has_view? view_name
         raise BlueprinterError, "View '#{view_name}' is not defined"
       end
-      prepared_object = select_columns(object, view_name: view_name)
+      prepared_object = Optimizer.new(object)
+                          .select(view_collection.fields_for(view))
       prepared_object = include_associations(prepared_object, view_name: view_name)
       if prepared_object.respond_to? :map
         prepared_object.map do |obj|
@@ -86,19 +88,6 @@ module Blueprinter
     end
     private_class_method :object_to_hash
 
-    def self.select_columns(object, view_name:)
-      unless defined?(ActiveRecord::Base) &&
-          object.is_a?(ActiveRecord::Base) &&
-          object.respond_to?(:klass)
-        return object
-      end
-      select_columns = (active_record_attributes(object) &
-        view_collection.fields_for(view).map(&:method)) +
-        required_lookup_attributes(object)
-      object.select(*select_columns)
-    end
-    private_class_method :select_columns
-
     def self.include_associations(object, view_name:)
       unless defined?(ActiveRecord::Base) &&
           object.is_a?(ActiveRecord::Base) &&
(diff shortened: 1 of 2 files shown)


lib/blueprinter/base.rb
```diff
@@ -42,7 +43,8 @@ def self.prepare(object, view:)
       unless view_collection.has_view? view_name
         raise BlueprinterError, "View '#{view_name}' is not defined"
       end
-      prepared_object = select_columns(object, view_name: view_name)
+      prepared_object = Optimizer.new(object)
+                          .select(view_collection.fields_for(view))
```
dlcarter: I think I would prefer to abstract this method to something like `optimize!` that then calls select, since we don't know at this point that this is an object that we should call `select` on. Plus, it allows us to expand the `Optimizer` later if we add more ways to optimize than just field selection

philipqnguyen: makes perfect sense!

lib/blueprinter/optimizer.rb
```diff
@@ -0,0 +1,36 @@
...
+                       required_lookup_attributes
+      object.select(*select_columns)
+    end
+
+    private
+
+    attr_reader :object
+
+    def active_record_relation?
+      !!defined?(ActiveRecord::Relation) &&
```
AllPurposeName: Any reason we're doing `!!defined?` instead of just `defined?`?

lib/blueprinter/base.rb
```diff
@@ -42,7 +43,8 @@ def self.prepare(object, view:)
       unless view_collection.has_view? view_name
         raise BlueprinterError, "View '#{view_name}' is not defined"
       end
-      prepared_object = select_columns(object, view_name: view_name)
+      prepared_object = Optimizer.new(object)
```
AllPurposeName: This makes more sense to me as a factory. I expect an optimizer being returned here.

lib/blueprinter/optimizer.rb
```diff
@@ -0,0 +1,36 @@
+module Blueprinter
+  class Optimizer
+    def initialize(object)
+      @object = object
+    end
+
+    def optimize!(fields)
+      return object unless active_record_relation?
```
AllPurposeName: In addition to my factory comment, this guard could be checked before initialization.

philipqnguyen: Good point. Though this will not be a problem after this class becomes a factory since we wont be initializing a new object.

lib/blueprinter/optimizer.rb
philipqnguyen: I think `defined?` is strange in that it does not return `true` or `false`. It either returns a String or nil. Generally methods with `?` should return a boolean.

If we do NOT force it into a boolean with `!!` and we are using Blueprinter on a non activerecord project, this method, `active_record_relation?`, would return `nil` instead of `false`.

philipqnguyen:
@dlcarter I'm rethinking about `Optimizer.optimize!`. Maybe it does not need a bang `!` because it is not altering any data in place. Does it make sense if i simply rename it to `Optimizer.optimize!`.

lib/blueprinter/optimizer.rb
AllPurposeName: Do we care about anything more than falsey?

AllPurposeName:
I don't think `Optimizer.optimize` is dangerous at all now, so I prefer to eschew the bang.

dlcarter:
Agree I think we can drop the ! since we're not modifying the object

Issue #26, opened 2017-10-10
Document the public methods via yard


philipqnguyen:
We should document all public methods via yard syntax. This way, if any developers ever come into trouble utilizing any of our public methods, they will have some reference when they look it up.

philipqnguyen:
merged #29

PR #29, opened 2017-10-13
Document methods

philipqnguyen:
- Adds yard inline docs to the public methods
- Mark private methods that are visible private. Marking them private warns devs not to use it as those methods could change any time and without warning.

The best way to see the docs is to:
1. bring the branch down locally.
2. `yard server -r`
3. Navigate to localhost:8808

#26

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ $ gem install blueprinter
 We use [Yard](https://yardoc.org/) for documentation. Here are the following
 documentation rules:
 
-- Document public all methods we expect to be utilized by the end developers.
+- Document all public methods we expect to be utilized by the end developers.
 - Methods that are not set to private due to ruby visibility rule limitations should be marked with `@api private`.
 
 ## Contributing
--- a/README.md
+++ b/README.md
@@ -23,6 +23,14 @@ Or install it yourself as:
 $ gem install blueprinter
 ```
 
+## Documentation
+
+We use [Yard](https://yardoc.org/) for documentation. Here are the following
+documentation rules:
+
+- Document public all methods we expect to be utilized by the end developers.
+- Methods that are not set to private due to ruby visibility rule limitations should be marked with `@api private`.
+
 ## Contributing
 Contribution directions go here.
 
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -13,10 +13,13 @@ module Blueprinter
     # Specify a field or method name used as an identifier. Usually, this is
     # something like :id
     #
+    # Note: identifiers are always rendered and considerered their own view,
+    # similar to the :default view.
+    #
     # @param method [Symbol] the method or field used as an identifier that you
     #   want to set for serialization.
     # @param name [Symbol] to rename the identifier key in the JSON
-    #   object. Defaults to method given.
+    #   output. Defaults to method given.
     # @param serializer [AssociationSerializer,PublicSendSerializer]
     #   Kind of serializer to use.
     #   Either define your own or use Blueprinter's premade serializers.
@@ -44,8 +47,8 @@ module Blueprinter
     #   Either define your own or use Blueprinter's premade serializers. The
     #   Default serializer is AssociationSerializer
     # @option options [Symbol] :name Use this to rename the method. Useful if
-    #   if you want your JSON key named differently than your object's field
-    #   or method name.
+    #   if you want your JSON key named differently in the output than your
+    #   object's field or method name.
     #
     # @example Specifying a user's first_name to be serialized.
     #   class UserBlueprint < Blueprinter::Base
@@ -66,7 +69,7 @@ module Blueprinter
     # @param method [Symbol] the association name
     # @param options [Hash] options to overide defaults.
     # @option options [Symbol] :name Use this to rename the association in the
-    #   JSON object.
+    #   JSON output.
     # @option options [Symbol] :view Specify the view to use or fall back to
     #   to the :default view.
     #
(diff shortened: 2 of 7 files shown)


mcclayton:
Nice! Looks great and works like a charm!

README.md
```diff
@@ -23,6 +23,14 @@ Or install it yourself as:
 $ gem install blueprinter
 ```
 
+## Documentation
+
+We use [Yard](https://yardoc.org/) for documentation. Here are the following
+documentation rules:
+
+- Document public all methods we expect to be utilized by the end developers.
```
mcclayton: `nb`: I think you mean `all public` not `public all`

philipqnguyen: nice catch

lib/blueprinter/base.rb
```diff
@@ -68,24 +128,86 @@ def self.prepare(object, view:)
...
     end
 
+
+    # Exclude a field that was mixed into the current view.
+    #
+    # @param field_name [Symbol] the field to exclude from the current view.
+    #
+    # @example Excluding a field from being included into the current view.
+    #   view :normal do
+    #     fields :position, company
```
AllPurposeName: Should :company be a symbol?

philipqnguyen: yes =D

lib/blueprinter/base.rb
```diff
@@ -10,18 +10,74 @@
 
 module Blueprinter
   class Base
+    # Specify a field or method name used as an identifier. Usually, this is
+    # something like :id
```
dlcarter: Add: `Note: identifiers are always rendered and considered their own view, similar to the :default view`

lib/blueprinter/base.rb
```diff
@@ -10,18 +10,74 @@
...
+    #
+    # @param method [Symbol] the method or field used as an identifier that you
+    #   want to set for serialization.
+    # @param name [Symbol] to rename the identifier key in the JSON
+    #   object. Defaults to method given.
```
dlcarter: change `object` to `output`

lib/blueprinter/base.rb
```diff
@@ -10,18 +10,74 @@
...
+    # @example Specifying a user's first_name to be serialized.
+    #   class UserBlueprint < Blueprinter::Base
+    #     field :first_name
+    #     # other code
+    #   end
+    #
+    # @return [Field] A Field object
     def self.field(method, options = {})
       name = options.delete(:name) || method
       serializer = options.delete(:serializer) || AssociationSerializer
```
dlcarter: NB: So your documentation drew attention to this - I think this should default to `PublicSendSerializer` vs `Association` - which is my fault but I don't know why I had it this way. I think we should clean up this logic in a separate PR?

lib/blueprinter/base.rb
```diff
@@ -10,18 +10,74 @@
...
+    #
+    # @param method [Symbol] the field or method name you want to include for
+    #   serialization.
+    # @param options [Hash] options to overide defaults.
+    # @option options [AssociationSerializer,PublicSendSerializer] :serializer
+    #   Kind of serializer to use.
+    #   Either define your own or use Blueprinter's premade serializers. The
+    #   Default serializer is AssociationSerializer
+    # @option options [Symbol] :name Use this to rename the method. Useful if
+    #   if you want your JSON key named differently than your object's field
```
dlcarter: changes in brackets: `named differently [in the output] than your...`

lib/blueprinter/base.rb
```diff
@@ -10,18 +10,74 @@
...
 
-    # Options:
-    #   preload: Set to false to stop trying to eagerly load associations
+    # Specify an associated object to be included for serialization.
+    # Takes a required method and an option.
+    #
+    # @param method [Symbol] the association name
+    # @param options [Hash] options to overide defaults.
+    # @option options [Symbol] :name Use this to rename the association in the
+    #   JSON object.
```
dlcarter: `object` -> `output`

lib/blueprinter/base.rb
```diff
@@ -68,24 +128,86 @@ def self.prepare(object, view:)
...
+    #   class UserBlueprint < Blueprinter::Base
+    #     # other code...
+    #     view :normal do
+    #       # some fields set here
+    #     end
+    #     view :extended do
+    #       include_view :normal # include fields specified from above.
+    #       field :description
+    #     end
+    #   end
```
dlcarter: We should probably augment this example to provide even more clarity - add maybe a `field :name` in the `:normal` view and then show how the output includes both `[:name, :description]`

lib/blueprinter/base.rb
```diff
@@ -68,24 +128,86 @@ def self.prepare(object, view:)
...
+    # Exclude a field that was mixed into the current view.
+    #
+    # @param field_name [Symbol] the field to exclude from the current view.
+    #
+    # @example Excluding a field from being included into the current view.
+    #   view :normal do
+    #     fields :position, company
+    #   end
+    #   view :special do
+    #     include_view :extended
```
dlcarter: `:extended` here should be `:normal`

lib/blueprinter/base.rb
```diff
@@ -68,24 +128,86 @@ def self.prepare(object, view:)
...
+    #     field :birthday
+    #     exclude :position
+    #   end
```
dlcarter: Again it would be nice to show the output of this logic - which should be `[:company, :birthday]`

lib/blueprinter/base.rb
philipqnguyen: Ok, issue created here #30

lib/blueprinter/base.rb
philipqnguyen: It's not showing here but I added it to this https://github.com/procore/blueprinter/pull/29/commits/c14c797aae8f0431834611fd24b0d58c1c2ac6ea

Issue #33, opened 2017-10-18
Allow custom options


philipqnguyen:
Add custom options that can be passed to render and utilized in the blueprint.

Custom options are any additional objects developers want to pass to the blueprint to help with their local methods.

Example use case:

```ruby
class UserBlueprint < Blueprinter::Base
  field :company_address
  def company_address
    @options[:company].address #something like this
  end
end

class UsersController
  def index
    # other code
    render json: Blueprinter.render(user, company: company)
  end
end
```

philipqnguyen:
This was resolved by #41

PR #38, opened 2017-10-19
Add local methods

### LINKED ISSUE #32: Local Method Support
We need to be able to allow end developers to add local methods in their blueprints.

For example: 

```ruby
  field :full_name

  def full_name
    "#{object.first_name} #{object.last_name}"
  end
end
```

philipqnguyen:
Resolves #32

~Add a method call `local_method`.  `local_method` accepts a method name and a method body. The method name is passed as an argument, and the method body is passed into a block.~

~Why we need something like this? There are situations where end devs want to define a method that are not used anywhere else except to have it serialized into JSON.~

Usage:
```ruby
  identifier :id
  fields :name
  local_method :age { 30 + 1 }
end
```

~So currently this works. But I also just now realized that it's use cases are currently narrow without first tackling #33. #33 would allow end developers to pass in custom options from `render` and utilize that in their blueprint class.~

Update: 10/20
~Now object is passed into the block so you can use it like so:~
``` ruby
  identifier :id
  fields :first_name
  local_method(:full_name) do |obj|
    "#{obj.first_name} #{obj.last_name}"
  end
end
```

Update 10/23
We allow the passing of a block to `field` like so:
```ruby
  # This would create a key value pair in the json output that looks like this:
  # {"full_name": "Jon Doe"}
  field(:full_name) do |obj|
    "#{obj.first_name} #{obj.last_name}"
  end
end
```

### DIFF
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -71,9 +71,11 @@ module Blueprinter
         {name: method, serializer: BlockSerializer, block: {method => block}}
       else
         {name: method, serializer: AssociationSerializer}
-      end.merge options
-      current_view << Field.new(method, options[:name], options[:serializer],
-        options)
+      end.merge(options)
+      current_view << Field.new(method,
+                                options[:name],
+                                options[:serializer],
+                                options)
     end
 
     # Specify an associated object to be included for serialization.
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -7,6 +7,7 @@ require_relative 'view_collection'
 require_relative 'optimizer'
 require_relative 'serializers/association_serializer'
 require_relative 'serializers/public_send_serializer'
+require_relative 'serializers/local_method_serializer'
 
 module Blueprinter
   class Base
@@ -217,6 +218,11 @@ module Blueprinter
       @current_view = view_collection[view_name]
       yield
       @current_view = view_collection[:default]
+    end
+
+    def self.local_method(method)
+      options = {local_methods: {method => yield}}
+      current_view << Field.new(method, method, LocalMethodSerializer, options)
     end
 
     private
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -231,8 +231,8 @@ module Blueprinter
     #   local_method(:age) { 31 }
     #
     # @return [Field] A Field object
-    def self.local_method(method)
-      options = {local_methods: {method => yield}}
+    def self.local_method(method, &block)
+      options = {local_methods: {method => block}}
       current_view << Field.new(method, method, LocalMethodSerializer, options)
     end
 
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -8,6 +8,7 @@ require_relative 'optimizer'
 require_relative 'serializers/association_serializer'
 require_relative 'serializers/public_send_serializer'
 require_relative 'serializers/local_method_serializer'
+require_relative 'serializers/block_serializer'
 
 module Blueprinter
   class Base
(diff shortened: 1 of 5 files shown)


lib/blueprinter/serializers/local_method_serializer.rb
```diff
@@ -0,0 +1,5 @@
+class Blueprinter::LocalMethodSerializer < Blueprinter::Serializer
+  def serialize(field_name, _object, options = {})
+    options[:local_methods][field_name]
```
dlcarter: If we change to this:
`options[:local_methods][field_name].call(_object, options)`

We should get the ability to interact with both the object AND any options passed into the blueprint, right?

philipqnguyen: that's a great idea

philipqnguyen: @dlcarter What do you see as useful for exposing the `options` variable through the block right now?

Right now, `render` does not accept an option args. 

We could allow `local_method` to accept `options` args, but do you see that as being useful right now? Or should that wait for #33 when we work on exposing options arguments for `render`?

philipqnguyen:
Just pushed a commit that would pass the object into the block. So can do this:

```ruby
      local_method(:full_name) do |obj|
        "#{obj.first_name} #{obj.last_name}"
      end
```

philipqnguyen:
We also should discuss more about this. Whether there is an alternative to calling it `local_method`, or if there is a better way all together.

AllPurposeName:
All we're attempting to do is define a field in a different way, so instead of the implied mapping between the field and the object, we're just intercepting and defining a field explicitly.

```ruby
field(:full_name) do |obj|
  "#{obj.first_name} #{obj.last_name}"
end
field(:age) { 30 + 1 }
```

philipqnguyen:
Really like that idea @AllPurposeName, I'll go ahead and update this pr with it.

lib/blueprinter/base.rb
```diff
@@ -49,18 +50,30 @@ def self.identifier(method, name: method, serializer: PublicSendSerializer)
...
+    #   end
+    #
     # @return [Field] A Field object
-    def self.field(method, options = {})
-      name = options.delete(:name) || method
-      serializer = options.delete(:serializer) || AssociationSerializer
-      current_view << Field.new(method, name, serializer, options)
+    def self.field(method, options = {}, &block)
+      options = if block_given?
+        {name: method, serializer: BlockSerializer, block: {method => block}}
```
AllPurposeName: NB: I suspect a FieldOptions class is yearning to be born.

lib/blueprinter/base.rb
```diff
@@ -49,18 +50,30 @@ def self.identifier(method, name: method, serializer: PublicSendSerializer)
...
+      else
+        {name: method, serializer: AssociationSerializer}
+      end.merge options
```
AllPurposeName: Would you mind putting parens here? 😛

lib/blueprinter/base.rb
```diff
@@ -49,18 +50,30 @@ def self.identifier(method, name: method, serializer: PublicSendSerializer)
...
+      current_view << Field.new(method, options[:name], options[:serializer],
+        options)
```
dlcarter: NB: The `options)` bugs me - would love to clean up the formatting of this line

lib/blueprinter/base.rb
philipqnguyen: ohhkaayy 😄

lib/blueprinter/base.rb
philipqnguyen: will do

trivedi:
@philipqnguyen is there any way to call a local method without defining a block? something similar to

```
field :age

def age
  18
end
```


the reason I ask is because sometimes the block can be large, so it's nice to break it into its own method. but a method that is only used in the serializer and not in the model (as mentioned in your PR summary).

philipqnguyen:
@trivedi

You can do 
```ruby
field :age do
  age
end

def self.age
  18
end
```

PR #39, opened 2017-10-24
Add readme for defining a field in the Blueprint

philipqnguyen:
This is the readme to describe the changes made from #38.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,34 @@ Output:
 }
 ```
 
+### Defining a field directly in the Blueprint
+
+You can define a field directly in the Blueprint by passing it a block. This is especially useful if the object does not already have such an attribute or method defined, and you want to define it specifically for use with the Blueprint. For example:
+
+```ruby
+class UserBlueprint < Blueprinter::Base
+  identifier :uuid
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -112,6 +112,30 @@ Output:
...
+
+```ruby
+class UserBlueprint < Blueprinter::Base
+  identifier :uuid
+  field(:full_name) do |user|
+    "#{user.first_name} #{user.last_name}"
+  end
+end
+
+puts UserBlueprint.render(user)
```
AllPurposeName: Maybe put this in the output? Additionally, it would be nice to copy paste from the terminal, so we have a 100% bonafide use case.

philipqnguyen: I could move the `puts UserBlueprint.render(user)` to a `usage:` section to stay consistent with the other examples.

Also could you clarify what you are expecting to see from the terminal that should be in the readme?

AllPurposeName: Hmm it seems the problem I have is with multiple different usages. Disregard my terminal comment. I'll follow up with my own PR. I do think making this portion of the docs similar to the others is worthwhile still 👍

PR #40, opened 2017-10-24
Allow viewing github markdown in development

philipqnguyen:
This will allow us to use github flavored markdown and view it during
development. Makes it especially useful for editing a readme locally.

Just run `yard server -r` and navigate to localhost:8808 and the
readme should now be in github flavored markdown.


philipqnguyen:
It's specified as development dependency so it won't be installed as part of `gem install blueprinter`

philipqnguyen:
closing this for now

PR #41, opened 2017-10-26
Render with options

philipqnguyen:
Allow `render` to take an options hash.
    
This `options` hash is internally called `local_options`, and it is
passed through to the serializer. 

I wanted to clearly distinguish options passed in through render, and the previously used options that was stored in Field. So i named the options passed in by render as `local_options`.
    
End devs can now pass an options hash to `render` and use it in the `field` block like this:
```ruby
  identifier :uuid
  field :company_id {|obj, options| options.company.id }
end

UserBlueprint.render(user, company: company)
#=> {"uuid": "x2334fw987a", "company_id": 2}
```

### DIFF
--- a/test/blueprinter_test.rb
+++ b/test/blueprinter_test.rb
@@ -29,7 +29,7 @@ class Blueprinter::Test < Minitest::Test
 
   def test_fields_using_custom_serializers
     upcase_serializer = Class.new(Blueprinter::Serializer) do
-      def serialize(field_name, object, options={})
+      def serialize(field_name, object, _local_options, _options={})
         object.public_send(field_name).upcase
       end
     end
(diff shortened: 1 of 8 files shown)


test/blueprinter_test.rb
```diff
@@ -91,13 +91,14 @@ def test_views
     )
   end
 
-  def test_field_with_block
-    my_obj = OpenStruct.new(id: 1, first_name: 'Meg', last_name: 'Ryan')
+  def test_render_with_options
```
AllPurposeName: This test has become confusing to me. We pass in a user and vehicle options to get back a serialized vehicle? What happened to `Meg Ryan` and why do we care about her vehicle?

Additionally, don't we want a test that the block works and a test that options work?

philipqnguyen: Thanks for catching this. I did not mean to over-write the block test with the options test. I meant to leave the `test_field_with_block` alone, and make a separate `test_render_with_options`, but i somehow ended up over-writing the former.

philipqnguyen:
I added the overwritten `test_render_with_block` back in there.

The 2.2.2 arg & block issue is not addressed here because I think @AllPurposeName you are going to address it?

AllPurposeName:
Yes, should be fixed in #48

PR #48, opened 2017-10-27
Conform tests to Ruby 2.2.2

AllPurposeName:
Prior to 2.4, Ruby does not handle mixing block arguments with regular arguments
WITHOUT parens correctly. Ideally, our DSL would look like:
```ruby
field :name { |obj| ... }
```
but <2.4 reads this like
```ruby
field(:name); { |obj ... }
```
finishing the method call before the block is even considered.


philipqnguyen:
Ohh circle ci will run the benchmark tests to which is normally excluded from the normal tests. I think those need update for 2.2.2 compatibility as well.

AllPurposeName:
Ah of course. On it.

AllPurposeName:
Ah of course. On it.

PR #49, opened 2017-10-30
Add tests for Blueprinter::View

AllPurposeName:
This class was untested. By adding tests I found a bug!

Now Views should properly reject duplicate fields being shoveled in.


lib/blueprinter/view.rb
```diff
@@ -1,29 +1,29 @@
 module Blueprinter
   # @api private
   class View
-    attr_reader :fields, :included_view_names, :excluded_field_names
+    attr_reader :excluded_field_names, :fields, :included_view_names, :name
 
-    def initialize(name)
+    def initialize(name, fields: {}, included_view_names: [], excluded_view_names: [])
```
philipqnguyen: NB: I notice this dependency injection being used a lot not just here, but also in other parts of blueprinter.
I think dependency injection makes sense if it's actually being used.

Otherwise, I generally don't think we should allow more arguments than what is actually being used.

Semantically, this is saying that we are requiring one argument and three optional arguments.
However, at no point in our code are we making use of those optional arguments. 

This can add confusion by letting other developers question whether adding just a `name` argument is enough, and whether they should do anything with the optional args.

I'm down to discuss more on this =)

dlcarter: Ah, this is actually a good catch, and I tend to agree. There doesn't seem to be much benefit from this particular code since it's not being used, and to Philip's point could engender developer confusion as to what this class does/expects

AllPurposeName: Since this is NB and it seems to be a difference of opinion, I've decided to go forward with the keyword argument additions.

In the end I'd rather trade a private class's ease of understanding for the ease of testing.

Issue #17, opened 2017-09-23
Implement OJ support


dlcarter:
Since Oj (https://github.com/ohler55/oj) is popular, performant, and above all FAST, we should either default to use it to serialize if detected, or provide a configuration to be able to set it to use Oj by default.

philipqnguyen:
If it's faster than the json gem, why don't we make it a required dependency?

philipqnguyen:
So after some discussion with @AllPurposeName and @dlcarter we have decided to by default use the json in Ruby standard library (not the gem extension) as the default JSON generator. We would build support for detecting (or configuring) JSON gem or OJ gem in the environment and use it if it exists. JSON gem and OJ gem would not be runtime dependencies in the gemspec.

philipqnguyen:
Oj is now supported and can be configured to be used with Blueprinter, see #54

PR #54, opened 2018-01-02
Configurable json generator

philipqnguyen:
The goal of this PR is to allow developers to utilize Oj to generate the JSON. This is to resolve issue #17. I would love some careful eyes on this, ideas, opinions.

You will notice that I removed the JSON gem dependency. That is done because the JSON gem and the subsequent `require 'json'` will override core ruby classes to add the `to_json` method. The override is global and it will affect everything in the developer's project. If the developer had already overridden classes with their own `to_json`, then we do not want to override that.

Instead, it is up to the developer to decide whether they want to use the JSON gem or Oj. Once they have installed and required the JSON gem or the Oj gem, they can configure Blueprinter to use whichever one of those:

```ruby
  config.generator = Oj # default is JSON
end
```

### DIFF
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -271,11 +271,7 @@ module Blueprinter
     private_class_method :include_associations
 
     def self.jsonify(blob)
-      if blob.respond_to? :to_json
-        blob.to_json
-      else
-        JSON.generate(blob)
-      end
+      Blueprinter.configuration.generator.generate(blob)
     end
     private_class_method :jsonify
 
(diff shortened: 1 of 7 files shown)


lib/blueprinter/base.rb
```diff
@@ -271,11 +270,7 @@ def self.include_associations(object, view_name:)
     private_class_method :include_associations
 
     def self.jsonify(blob)
-      if blob.respond_to? :to_json
-        blob.to_json
-      else
-        JSON.generate(blob)
-      end
+      Blueprinter.configuration.generator.generate(blob)
```
philipqnguyen: This is where the configured generator is called to generate.

PR #55, opened 2018-01-05
Rspec

philipqnguyen:
I added rspec so that we can move out of minitest. I think Rspec is easier to work with, and it has some nice re-usability features. In the end it really comes down to preference to be honest. Note that rspec would only be installed if working with this gem from a development perspective and not as part of the normal gem install process.

There are a lot of changes, but 99% of the changes are just renaming and moving files. To help reviewers focus, this b895a9874a651af1f878603082890dcd8dbc0ff5 contains ONLY the differences between the previous minitest tests and the new rspec specs. Since both the active_record_test.rb and the blueprinter_test.rb were testing for the same things just under a different context (one with Rails and one without Rails), I used shared examples so that they both run the same spec examples but with one under a Rails context and one outside of Rails context

I tried to meaningfully use the `describe`, `context`, and `it` blocks so that the output of the specs would explain the capabilities of blueprinter. Here is the rspec output:

```
::View
  #include_view(:view_name)
    should return [:view_name]
    should set #included_view_names to [:view_name]
  #exclude_field(:view_name)
    should return [:view_name]
    should set #excluded_field_names to [:view_name]
  #<<(field)
    Given a field that does not exist
      should return field
      should set #fields to {field.name => field}
    Given a field that already exists
      should raise Blueprinter::BlueprinterError
      should not set #fields
  #fields
    Given no fields
      should eq {}
    Given existing fields
      should eq {field.name => field}

::Base
  ::render
    Outside Rails project
      Given blueprint has ::field
        returns json with specified fields
      Given blueprint has ::fields
        returns json with specified fields
      Given blueprint has ::field with a :name argument
        returns json with a renamed field
      Given blueprint has ::field with a :serializer argument
        returns json derived from a custom serializer
      Given blueprint has ::view
        returns json derived from a view
      Given blueprint has ::field with a block
        returns json with values derived from a block
      Given ::render with options
        returns json with values derived from options
    Inside Rails project
      Given blueprint has ::field
        returns json with specified fields
      Given blueprint has ::fields
        returns json with specified fields
      Given blueprint has ::field with a :name argument
        returns json with a renamed field
      Given blueprint has ::field with a :serializer argument
        returns json derived from a custom serializer
      Given blueprint has ::view
        returns json derived from a view
      Given blueprint has ::field with a block
        returns json with values derived from a block
      Given ::render with options
        returns json with values derived from options
```


spec/spec_helper.rb
```diff
@@ -0,0 +1,104 @@
+# This file was generated by the `rspec --init` command. Conventionally, all
```
AllPurposeName: NB: Could we remove this comment nonsense?

PR #56, opened 2018-01-10
Support hash object passed to render

### LINKED ISSUE #43: `field.serializer.serialize`
https://github.com/procore/blueprinter/blob/e4c5daa77e869dea518618a74e058ca78d46aada/lib/blueprinter/base.rb#L241

This `Base` class seems to know too much about the internals of `Field`. Does that need to be moved into the `Field` object? So just call `field.serialize(arg1, arg2)`

philipqnguyen:
Support Hash by using AutoSerializer as default

AutoSerializer will delegate between HashSerializer to serialize
Hash objects or to PublicSendSerializer to serialize objects with
dot notation attributes.

We also don't iterate over all objects that respond to `:map` anymore
because `Hash` responds to `:map`, but we wouldn't want that. So we
have to be more specific on what objects we treat like an array.
    
In the process of creating new serializers, I noticed that the Base class seems to know about Field and the Serializer class way too much. So I move the responsibility of invoking
the serializer to the Field. This actually resolves #43.

### DIFF
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -244,10 +244,7 @@ module Blueprinter
 
     def self.object_to_hash(object, view_name:, local_options:)
       view_collection.fields_for(view_name).each_with_object({}) do |field, hash|
-        hash[field.name] = field.serializer.serialize(field.method,
-                                                      object,
-                                                      local_options,
-                                                      field.options)
+        hash[field.name] = field.serialize(object, local_options)
       end
     end
     private_class_method :object_to_hash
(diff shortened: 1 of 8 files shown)


philipqnguyen:
@dj yes, I have been thinking about it. Haven't quite thought of a good solution. I do think we need to normalize user input soon. I didn't want to deal with that in this PR yet because I just wanted to focus this PR more on supporting Hash.

dj:
Hello, I think you have me confused with someone else.

AllPurposeName:
Welcome, nice name 😉

PR #59, opened 2018-01-18
`required_lookup_attributes_for` could be nil

philipqnguyen:
So we need to remove the nil cases.


philipqnguyen:
maybe wont need this, depending on how we go with #60

PR #60, opened 2018-01-19
Remove broken optimizer fix association

philipqnguyen:
Two things:
1. Remove Optimizer
2. Fix assocations
----
    Optimizer is not currently working and is causing multiple problems.
    Until we come up with a solution, I think it is best to remove
    Optimizer so that we can start accepting ActiveRecord objects that
    have relations with other objects.
    We should re-introduce Optimizer at a later time when we have a working
    solution.

    To fix association, we require that association accept a blueprint so that we can serialize
    associated objects with the correct blueprint.


AllPurposeName:
I feel like these two changes are quite distinct. Could we (at least going forward) break these things out into multiple PRs?

lib/blueprinter/serializers/association_serializer.rb
```diff
@@ -1,10 +1,8 @@
...
-    if options[:blueprint]
-      view = options[:view] || :default
-      options[:blueprint].prepare(object.public_send(association_name), view_name: view, local_options: local_options)
-    else
-      object.public_send(association_name)
-    end
+    value = object.public_send(association_name)
+    return value if value.nil?
+    view = options[:view] || :default
+    options[:blueprint].prepare(value, view_name: view, local_options: local_options)
```
dlcarter: So, I'm not sure if you intended this or not, but a side effect of this code is that it makes it so that you now always *have* to have a Blueprint for every association. I think that could be considered a reasonable tradeoff, but I want to make sure we're explicitly making that decision.

philipqnguyen: Ya, so I do not understand how it would work if the associated object does not have a blueprint.

philipqnguyen:
@AllPurposeName Yes, I can separate the removal of Optimizer into another PR.

philipqnguyen:
removal of optimizer has been cherry-picked out into #61

lib/blueprinter/serializers/association_serializer.rb
AllPurposeName: I am for having the requirement that any object being serialized within this system must use a blueprint... I've personally been bitten by having both AMS and jbuilder for the same resource far too many times to try to fashion some coexistence between any of them.

spec/integrations/base_spec.rb
```diff
@@ -67,6 +67,40 @@
...
+          let(:blueprint) do
+            vehicle_blueprint = Class.new(Blueprinter::Base) do
+              fields :make
+            end
+            Class.new(Blueprinter::Base) do
+              identifier :id
+              association :vehicles, blueprint: vehicle_blueprint
+            end
+          end
+          it('returns json with association') { should eq(result) }
```
AllPurposeName: This is so implicit and confusing. Maybe we should switch to MiniTest 😜

philipqnguyen: ok i admin i over-utilized some of rspec's "features" that does make it quite implicit. I'll make more explicit specs going forward =D

PR #63, opened 2018-02-11
Change 'block.call to yield'

NoobCIT:
Past benchmark tests have shown that block.call is almost two times
slower than yield in older versions of Ruby.


spec/benchmarks/benchmark_helper.rb
```diff
@@ -7,7 +7,7 @@ def iterate(&block)
     start = Time.now
     count = 0
     while Time.now - start <= 1.second do
-      block.call
+      yield
```
philipqnguyen: Would you remove the `(&block)` argument in line 6? `yield` would not need that argument.

PR #64, opened 2018-02-13
[Security] Upgrade nokogiri from 1.8.0 to 1.8.2

philipqnguyen:
There is a security vulnerability in Nokogiri 1.8.0, so we upgrade it
to 1.8.2 (latest).

In truth, the vulnerabilty never affected Blueprinter because Nokogiri
was only used for the dummy rails app when running tests. Blueprinter
itself never depended on Nokogiri. However we should upgrade it any
ways in case somebody uses the dummy app in unexpected ways.


mcclayton:
I'm thinking the Gemfile.lock is kind of an odd place to upgrade this version. I think it may be a little more explicit to pin the version in the blueprinter.gemspec. Somewhat irrelevant since this is only a dev dependency and not a real security issue - but just my two cents.

Issue #67, opened 2018-03-19
Rename Blueprint.render


philipqnguyen:
`Blueprint.render` sounds weird when using it inside a Rails controller because there would be two `render` methods being used, one on the controller and one on the blueprint. See below:

```ruby
def show
  render json: PostBlueprinter.render(@post)
end
```

`Blueprint.print` is a better method name, and it should alias to `Blueprint.render`.

philipqnguyen:
closing this because we are not sure if we should rename it anymore.

PR #71, opened 2018-03-22
Provides alias to reduce method name repetitions per issue #67

NoobCIT:
I used a metaclass to provide an alias for the render method per issue #67.

### DIFF
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -131,6 +131,10 @@ module Blueprinter
       jsonify(prepare(object, view_name: view_name, local_options: options))
     end
 
+    class << self
+      alias_method :print, :render
+    end
+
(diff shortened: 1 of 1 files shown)


philipqnguyen:
@NoobCIT thanks for submitting this PR.

So originally #67 did say to rename from `render` to `print`, but we might have to rename it to something else again. The problem is that `print` is already a built-in Ruby method, and so it could be confusing to some people. Right now we're trying to come up with some ideas for naming. If you have a naming idea, feel free to comment with it.

NoobCIT:
Here are some ideas: deliver, process, out, output, stamp, log, issue, imprint.

philipqnguyen:
closing this for now because we are not sure anymore if a name change is needed. Closes #67

PR #72, opened 2018-03-23
Rename Serializer classes to Extractor

philipqnguyen:
I think that Serializer doesn't identify what these classes were
doing. These classes are extracting values based on how they were
stored in the input object. So I think it's better to call it
Extractor.

I'm up to changing it to another name if Extractor doesn't sound
correct.


wai-procore:
👍

PR #74, opened 2018-04-23
Changed FactoryGirl to FactoryBot

das3in:
Thoughtbot has deprecated the FactoryGirl gem in favor of FactoryBot.


philipqnguyen:
Thanks! This is awesome.

Issue #70, opened 2018-03-19
Serialize datetime field


philipqnguyen:
How should Blueprinter serialize datetime fields?

Should Blueprinter automatically check each fields to see if it is a DateTime or Time object?
Or should we provide a way for the user to specify that a field is of DateTime or Time? 

Either way, we also would need to allow users to customize the datetime format that they would like to serialize to.

psloomis:
How about this:
```ruby
class InsuranceBlueprint < Blueprinter::Base
  field :expiration_date, date_format: "%d %m %Y"
end
```
If `date_format` is passed as an option, we just assume the object is a `DateTime` or `Time` or `Date`.

philipqnguyen:
I definitely like that idea @psloomis would you like to take a stab at a PR for that?

mcclayton:
Hi @anlek, I believe the possibility of having a globally set datetime format was being discussed in issue: https://github.com/procore/blueprinter/issues/120

PR #84, opened 2018-05-30
Add custom date and time formatting

njbbaer:
Blueprinter needs a mechanism to specify custom serialization for Date and DateTime fields.

To define a custom format for a Date or DateTime field, include the option `datetime_format` with the associated `strptime` format.

Usage:
```ruby
  identifier :name
end
```

Output:
```json
{
  "name": "John Doe",
  "birthday": "03/04/1994"
}
```

### DIFF
--- a/lib/blueprinter/extractors/auto_extractor.rb
+++ b/lib/blueprinter/extractors/auto_extractor.rb
@@ -3,7 +3,15 @@ module Blueprinter
     def extract(field_name, object, local_options, options = {})
       extractor = object.is_a?(Hash) ? HashExtractor : PublicSendExtractor
       extraction = extractor.extract(field_name, object, local_options, options)
-      options.key?(:datetime_format) ? extraction.strftime(options[:datetime_format]) : extraction
+      options.key?(:datetime_format) ? format_datetime(extraction, options[:datetime_format]) : extraction
+    end
+
+    private
+
+    def format_datetime(datetime, format)
+      datetime.strftime(format)
+    rescue NoMethodError
+      raise BlueprinterError, 'Cannot format invalid DateTime object'
     end
   end
 end
--- a/lib/blueprinter/extractors/auto_extractor.rb
+++ b/lib/blueprinter/extractors/auto_extractor.rb
@@ -1,15 +1,9 @@
 module Blueprinter
   class AutoExtractor < Extractor
     def extract(field_name, object, local_options, options = {})
-      extractor =
-        if object.is_a?(Hash)
-          HashExtractor
-        elsif options.key?(:datetime_format)
-          DateTimeExtractor
-        else
-          PublicSendExtractor
-        end
-      extractor.extract(field_name, object, local_options, options)
+      extractor = object.is_a?(Hash) ? HashExtractor : PublicSendExtractor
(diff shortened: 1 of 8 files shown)


njbbaer:
Resolves #70

lib/blueprinter/extractors/auto_extractor.rb
```diff
@@ -2,7 +2,8 @@ module Blueprinter
   class AutoExtractor < Extractor
     def extract(field_name, object, local_options, options = {})
       extractor = object.is_a?(Hash) ? HashExtractor : PublicSendExtractor
-      extractor.extract(field_name, object, local_options, options)
+      extraction = extractor.extract(field_name, object, local_options, options)
+      options.key?(:datetime_format) ? extraction.strftime(options[:datetime_format]) : extraction
```
mcclayton: What about when extraction is not a DateTime object?
Do we want to surface our own error in that case? i.e. something like: `"Attempting to format invalid DateTime object"`

njbbaer: Makes sense. Done 👍

lib/blueprinter/extractors/auto_extractor.rb
```diff
@@ -2,7 +2,16 @@ module Blueprinter
   class AutoExtractor < Extractor
     def extract(field_name, object, local_options, options = {})
       extractor = object.is_a?(Hash) ? HashExtractor : PublicSendExtractor
-      extractor.extract(field_name, object, local_options, options)
+      extraction = extractor.extract(field_name, object, local_options, options)
+      options.key?(:datetime_format) ? format_datetime(extraction, options[:datetime_format]) : extraction
```
philipqnguyen: I'm just wondering if this formatting should be here or it should be in it's own extractor or maybe an entirely new class??

njbbaer: Originally I had it as its own extractor, but it has to work for hashes as well and I didn't want to avoid the existing `HashExtractor` logic.

philipqnguyen: I'm just thinking where best to put this. Now that `AutoExtractor` is both extracting and formatting.

Maybe formatting dates isnt an `Extractor` because it's not extracting. Maybe it could be a kind of helper module called `FormatHelpers` or a class called `Formatter` where you could do something like `Formatter.new(extraction, option).format`

philipqnguyen:
Would you add reference docs here like this: https://github.com/procore/blueprinter/blame/master/lib/blueprinter/base.rb#L53-L55

That's what generates these references: https://www.rubydoc.info/gems/blueprinter/Blueprinter/Base#field-class_method

lib/blueprinter/extractors/auto_extractor.rb
AllPurposeName: I can really get behind a `Formatter` class with a constructor API of something like `Formatter.format(extraction, options)`. Let it do it's own initialization if necessary 👍

njbbaer:
Added reference docs @philipqnguyen

philipqnguyen:
released in 0.6.0

PR #86, opened 2018-06-01
Add conditional fields support

ojab:
See https://github.com/procore/blueprinter/issues/81
One omission: methods should be declared before field declaration using them, i. e.
```ruby
class UserBlueprinter < Blueprinter::Base
  # field :first_name, if: :my_method here will fail
  def my_method(user, options)
    # some code
  end

  field :first_name, if: :my_method
end
```
because methods are not available yet at field declaration time and I'm resolving Symbol to Method there. Is it worth fixing or I should just add a note to README?

Other one: AFAICS collection serialization is not tested at the moment, so I also skipped it.

### DIFF
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -37,7 +37,7 @@ module Blueprinter
     #
     # @return [Field] A Field object
     def self.identifier(method, name: method, extractor: AutoExtractor)
-      view_collection[:identifier] << Field.new(method, name, extractor)
+      view_collection[:identifier] << Field.new(method, name, extractor, self)
     end
 
     # Specify a field or method name to be included for serialization.
@@ -53,6 +53,14 @@ module Blueprinter
     # @option options [Symbol] :name Use this to rename the method. Useful if
     #   if you want your JSON key named differently in the output than your
     #   object's field or method name.
+    # @option options [Symbol,Proc] :if Specifies a method, proc or string to
+    #   call to determine if the field should be included (e.g.
+    #   `if: :include_first_name?, or if: Proc.new { |user, options| options[:current_user] == user }).
+    #   The method, proc or string should return or evaluate to a true or false value.
+    # @option options [Symbol,Proc] :unless Specifies a method, proc or string
+    #   to call to determine if the field should be included (e.g.
+    #   `unless: :include_first_name?, or unless: Proc.new { |user, options| options[:current_user] != user }).
+    #   The method, proc or string should return or evaluate to a true or false value.
     # @yield [Object] The object passed to `render` is also passed to the
     #   block.
     #
@@ -68,6 +76,17 @@ module Blueprinter
     #     # other code
     #   end
     #
+    # @example Passing an if proc and unless method..
+    #   class UserBlueprint < Blueprinter::Base
+    #     def skip_first_name?(user, options)
+    #       user.first_name == options[:first_name]
+    #     end
+    #
+    #     field :first_name, unless: :skip_first_name?
+    #     field :last_name, if: ->(user, options) { user.first_name != options[:first_name] }
+    #     # other code
+    #   end
+    #
     # @return [Field] A Field object
     def self.field(method, options = {}, &block)
       options = if block_given?
@@ -78,6 +97,7 @@ module Blueprinter
       current_view << Field.new(method,
                                 options[:name],
                                 options[:extractor],
+                                self,
                                 options)
     end
 
@@ -107,6 +127,7 @@ module Blueprinter
       current_view << Field.new(method,
                                        name,
                                        AssociationExtractor,
+                                       self,
                                        options.merge(association: true))
     end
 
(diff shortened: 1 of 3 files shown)


lib/blueprinter/base.rb
```diff
@@ -68,13 +76,29 @@ def self.identifier(method, name: method, extractor: AutoExtractor)
...
       options = if block_given?
         {name: method, extractor: BlockExtractor, block: {method => block}}
       else
         {name: method, extractor: AutoExtractor}
       end.merge(options)
+      %i[if unless].each do |condition|
+        if options[condition].is_a?(Symbol)
+          options[condition] = self.method(options.fetch(condition))
+        end
+      end
```
philipqnguyen: There might be a way to resolve the conditional method at `render` execution time rather than at `field` declaration time, allowing conditional methods to be defined anywhere in the blueprint. Here's the idea, let me know what you think =)

What if you move this chunk of code into the `skip?` method you made in `Field`?

lib/blueprinter/base.rb
```diff
@@ -270,6 +294,7 @@ def self.view(view_name)
 
     def self.object_to_hash(object, view_name:, local_options:)
       view_collection.fields_for(view_name).each_with_object({}) do |field, hash|
+        next if field.skip?(object, local_options)
```
philipqnguyen: We would need to pass `self` here. `field.skip?(object, local_options, self)`

lib/blueprinter/field.rb
```diff
@@ -11,4 +11,14 @@ def initialize(method, name, extractor, options = {})
   def extract(object, local_options)
     extractor.extract(method, object, local_options, options)
   end
+
+  def skip?(object, local_options)
```
philipqnguyen: Argument list would be altered to look like: `skip?(object, local_options, blueprint)`

Then you can place your above code here in this method:
```
    %i[if unless].each do |condition|
      if options[condition].is_a?(Symbol)
        options[condition] = blueprint.method(options.fetch(condition))
      end
    end
```

philipqnguyen:
And as for collection testing... yes we need to get that in there. It would be in a separate PR though.

ojab:
Passing blueprint to `Field` at initialization time looks more clear to me. 
Pushed that version, what do you think?

spec/integrations/shared/base_render_examples.rb
```diff
@@ -50,6 +50,74 @@ def extract(field_name, object, _local_options, _options={})
...
+        context 'and no :unless argument' do
+          if value
+            it { should eq(result_with_first_name) }
+          else
+            it { should eq(result_without_first_name) }
+          end
+        end
+
+        variants.each do |other_type, other_value|
+          context "and :unless is #{other_type} returning #{other_value}" do
```
philipqnguyen: The current location of this context block makes it look like `:unless` depends on first having `:if` passed to `field`.  Both `:if` and `:unless` should be at the same level of context right?

I'm thinking of something like:

```
    Inside Rails project
      Given blueprint has ::field with a conditional argument
        Given the conditional is :if
           ... other contexts / specs
        Given the conditional is :unless
           ... other contexts / spces
    Outside Rails project
      ... repeat from above ...
```

spec/integrations/shared/base_render_examples.rb
```diff
@@ -50,6 +50,74 @@ def extract(field_name, object, _local_options, _options={})
...
+            let(:unless_value) { other_value }
+            before do
+              field_options[:unless] = if type == :method then :unless_method
+                                       else unless_proc
+                                       end
+            end
+
+            if value && !other_value
+              it { should eq(result_with_first_name) }
```
philipqnguyen: The current rspec printout out for these `it` blocks are something like:
```
and :unless is proc returning false
  should eq "{\"first_name\":\"Meg\",\"id\":15}"
```
or
```
and :unless is proc returning true
  should eq "{\"id\":16}"`
```

When reading the rspec printout, we don't actually care whether `first_name` is there or not, as that is just the example. What we care is whether the field with the conditional is there or not.

I think a description for the `it` block should be there to clarify the printout. 
`it ('does not serialize the conditional field') { should eq(result_without_first_name) }`
or
`it ('serializes the conditional field') { should eq(result_with_first_name) }`

which gives a printout of something like:

```
and :unless is proc returning true
  serialize the conditional field
```
or
```
and :unless is proc returning false
  does not serialize the conditional field
```

This comment applies to the other `it` blocks in this PR as well.

spec/integrations/shared/base_render_examples.rb
ojab: it actually is: it's specs for both :if & :unless passed. Added specs for :unless only clarity and fixed examples descriptions, now it looks like:
```
    Inside Rails project
      Given blueprint has ::field with a conditional argument
        Given the conditional is :if proc returning true
          and no :unless conditional
            serializes the conditional field
          and :unless conditional is proc returning true
            does not serialize the conditional field
          …
        Given the conditional is :unless proc returning true and no :if conditional
          does not serialize the conditional field
        Given the conditional is :if proc returning false
          and no :unless conditional
            does not serialize the conditional field
          and :unless conditional is proc returning true
          …
```

philipqnguyen:
released in 0.6.0

Issue #92, opened 2018-08-30
Idea : provide a mapper


Systho:
I've had a little time to work on my problem described in #90 and I made a better solution (at leats to my eyes )

Here is the code of my patch (which I execute at the start of my app)

```ruby
module BlueprintDelegation

  def mapping(&block)
    delegator_class.class_eval(&block)
  end

  #below this is private API

  # this is the tricky part, this method must store a different delegator class in each blueprint
  def delegator_class
    @delegator_class ||= Class.new(SimpleDelegator)
  end

  def build_object_delegator(object)
    delegator_class.new(object)
  end
  
  def object_to_hash(object, **named_params)
    object_delegator = build_object_delegator(object)
    super(object_delegator, **named_params)
  end
  
end

Blueprinter::Base.singleton_class.prepend BlueprintDelegation

```

The usage is done like this :

```ruby
module PublicApi
  module V1
    class UserBlueprint < BaseBlueprint

      field :first_name
      field :last_name
      field :email

      view(:show) do
        association(:participations, blueprint: ParticipationBlueprint)
      end

      mapping do
        def participations
          learner.participations
        end
      end

    end
  end
end
```

I feel this is prettier than passing blocks all over the place and is usable with fields and associations and any extractor .

If you like the idea I'll make a PR

philipqnguyen:
@Systho I like the idea because this would suddenly allow `fields` and `associations` the ability to customize, but this also does what `field` blocks already does.

If we introduce `mapping`, should we continue supporting `field` blocks or should it be deprecated? For me, I don't like the idea of officially supporting both `field` blocks and `mapping` side by side because it can be confusing for people to know when to use which.

In addition, `field` blocks supports additional properties option https://github.com/procore/blueprinter#passing-additional-properties-to-render
Should `mapping` handle that too?

@dlcarter do you want to chime in?

Systho:
I think the `mapping` idea is better than the block because : 
 - it enables fields and associations to have mapping without duplicating code.
 - Since it's usual OO, we can use the usual design tools to improve it. Blocks have fewer improvement possibilities.

I think you can either deprecate blocks, or have the provided block define a method in the delegator_class :) 

I can make `options` available in the mapping methods as a readable_attribute of the delegator_class.

philipqnguyen:
[...]

PR #93, opened 2018-09-06
Allow blueprint to define a mapping for the rendered object

Systho:
PR related to  #92

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ spec/dummy/log/*.log
 spec/dummy/tmp/
 doc/
 .yardoc/
+.idea
+.tool-versions
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -7,6 +7,7 @@ require_relative 'extractors/block_extractor'
 require_relative 'extractors/hash_extractor'
 require_relative 'extractors/public_send_extractor'
 require_relative 'field'
+require_relative 'mapper'
 require_relative 'view'
 require_relative 'view_collection'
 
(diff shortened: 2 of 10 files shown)


philipqnguyen:
The specs are failing because circle ci runs the specs at ruby 2.2.2 which is the minimum supported ruby version for blueprinter. You might want to add `require 'delegate'` since it doesn't seem like it is automatically loaded in 2.2.x

vinaya-procore:
This is a nice idea!

But I think it will create more confusion, as `Blueprinter` will end up having `field`, `fields`, and `mapping` to choose from. I see the convenience that `mapping` block brings, enabling multiple methods to be grouped together. But I do not really feel that it provides better value than what `field` already offers, grouping things that do not necessarily belong together.

I recommend that we stick to using `field` as a block when serializing an association that's not directly related to the object.

Thanks.

Systho:
@vinaya-procore  actually the point is not to group multiple methods together but to be able to have an association return a collection which is not immediately on the object.

The initial idea was to allow `association` to receive block (like `field` do ) but when I wrote that implementation I realised that it brought some duplication which was hiding the concept of an mapper object.

I can rewrite the DSL to simply allow passing block to association but I would really suggest that technically those blocks are just adding methods to an invisible wrapper around the rendered object.

philipqnguyen:
@Systho we can accept a PR that adds block to `association` if you're still up for it, but `mapping` is too much of a change to the public interface.

philipqnguyen:
@Systho you're right that there isn't a clear path to storing blocks. Currently `Field` stores an instance of the BlockExtractor and the block itself, which may or may not be the best way. One of the focus I'm hoping to tackle in the foreseeable future is to cut down on time spent inside blueprinter when serializing ruby objects. SimpleDelegator uses `method_missing` to match the method calls, which can be slow. 

I ran a quick ruby-prof test on your branch:

```ruby
# Using a blueprint like this
class ProfilerBlueprint < Blueprinter::Base
  identifier :id
  fields :attr_1,
         :attr_2,
         :attr_4,
         :attr_5,
         :attr_6,
         :attr_7,
         :attr_8,
         :attr_9,
         :attr_10,
         :attr_11,
         :attr_12,
         :attr_13,
         :attr_14,
         :attr_15,
         :attr_16,
         :attr_17,
         :attr_18,
         :attr_19,
         :attr_20,
         :attr_21
  view :normal do
    fields :attr_22, :attr_25
    field :attr_23, name: :renamed_attr
  end
  view :extended do
    include_view :normal
    field :attr_24
  end
  view :special do
    include_view :extended
    exclude :attr_25
  end
end

objects = 100.times.map {obj}
require 'ruby-prof'
RubyProf.measure_mode = RubyProf::PROCESS_TIME
RubyProf.start
ProfilerBlueprint.render(objects, view: :special)
result = RubyProf.stop
printer = RubyProf::FlatPrinter.new(result)
printer.print(STDOUT)
```

this branch process time: 0.359882
master branch process time: 0.225519

Obviously those results comes from my machine so the exact number will be different, but I think the difference is pretty significant. If the decision is between speed or using a delegator, I will have to pick speed, considering that speed is an important aspect of a serializer.

philipqnguyen:
Closing this PR which also closes #92

reopen the issue if you wish for further discussion.

PR #97, opened 2018-10-03
Remove Gemfile.lock and move dependencies to gemspec

hugopeixoto:
Hi there! Let me know if the following changes make sense:

From what I understand, it's usually not recommended to commit `Gemfile.lock`[1].
This PR removes it and adds it to gitignore.

Additionally, a dev dependency was added to the `Gemfile` instead of being added to the `gemspec`.
This PR also moves that dependency.

[1] https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/


blueprinter.gemspec
```diff
@@ -25,4 +25,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency "rspec", "~> 3.7"
   s.add_development_dependency "sqlite3"
   s.add_development_dependency "yard", "~> 0.9.11"
+  s.add_development_dependency "factory_bot"
```
philipqnguyen: would you move this up to retain ABC ordering?

hugopeixoto: will do, I wasn't aware of the ordering!

hugopeixoto:
I was trying to figure out why factorybot wasn't being detected. Kind of surprised it works locally, though. I'll add the require, thanks for the pointer

Issue #98, opened 2018-10-08
Ruby version / circleci 2.0


hugopeixoto:
Working on upgrading the circleci config to 2.0, I noted that the required ruby version is `2.2.2`, which is not supported any more.

Should this be bumped to 2.3, or even 2.4? Not sure what the backwards compatibility policy is.

Related: Should I test for multiple ruby versions in circleci? Right now I have a working config that runs on 2.3.7, 2.4.4 and 2.5.1, but this may be overkill.

PR #99, opened 2018-10-09
Upgrade to circleci 2.0

hugopeixoto:
I've upgraded to circleci 2.0 and included tests for multiple ruby versions.

I'm having some issues when running the benchmark tests. Ruby 2.2 tends to fail the linear test (~98%). We might need to increase the threshold to make this pass.


philipqnguyen:
Thanks for tackling #98 it looks good!

Yes let's bring the threshold to 98, we can adjust it again if it still becomes a problem at 98 in future tests.

hugopeixoto:
I reduced it to 98%, should be ready to go

PR #102, opened 2018-10-11
Remove rails app from spec directory

hugopeixoto:
Instead of depending on a full rails app to setup a database, migrations
and models, configure ActiveRecord in spec/activerecord_helper.rb using
an sqlite3 in memory database.

By removing the rails app, autoloading was disabled when running
specs/benchmarks. This uncovered some missing requires on both lib/ and
spec/.


lib/blueprinter/configuration.rb
```diff
@@ -1,3 +1,5 @@
+require 'json'
```
philipqnguyen: This has to be removed. We purposefully did not `require 'json'` because we do not want to assume that people will be using the `json` lib. The `json` lib is very invasive, it will override basic ruby objects and add `to_json` methods to them.

It is up to users of blueprinter to require the json marshaller gems of their choosing. It's noted here: https://github.com/procore/blueprinter#installation

lib/blueprinter/base.rb
```diff
@@ -1,12 +1,13 @@
 require_relative 'blueprinter_error'
-require_relative 'helpers/active_record_helpers'
+require_relative 'configuration'
```
philipqnguyen: This is already required in https://github.com/procore/blueprinter/blame/master/lib/blueprinter.rb#L1 so this is not needed.

spec/activerecord_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+    t.string "model"
+    t.integer "miles"
+    t.integer "user_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["user_id"], name: "index_vehicles_on_user_id"
+  end
+end
+
+ActiveRecord::Migration.maintain_test_schema!
```
philipqnguyen: This is great!

lib/blueprinter/configuration.rb
philipqnguyen: You should just need to `require 'json'` in the spec_helper only for the specs.

lib/blueprinter/base.rb
hugopeixoto: This means that we cannot require `blueprinter/base.rb`, since it uses `Blueprinter.configuration`, without requiring `blueprinter.rb`. Is this the intended behavior? Should the `configuration` require be only in base, insetad of in `blueprinter.rb`?

philipqnguyen: I see your point. Yea we can leave configuration here, just remove it from blueprinter.rb and you're good to go!

PR #103, opened 2018-10-11
Circleci ruby test

philipqnguyen:
Ruby 2.2.0 has severe memory leak problems, which could be the reason
why it is regularly failing the linear benchmark test.

See https://bugs.ruby-lang.org/issues/10686

The next closest 2.2.x official ruby docker image is 2.2.10, so using
2.2.10 instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,10 +11,10 @@ steps: &steps_spec
       command: bundle exec rake benchmarks
 
 jobs:
-  ruby-2.2:
+  ruby-2.2.10:
     steps: *steps_spec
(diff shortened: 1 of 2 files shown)


hugopeixoto:
The `ruby:2.2` docker image contains the latest 2.2.x ruby version:

```
$ docker run -ti ruby:2.2 ruby -e "puts RUBY_VERSION"
2.2.10
```

So I think your results might be just because benchmarks always fluctuate a bit

philipqnguyen:
@hugopeixoto didn't realize thanks.

Issue #80, opened 2018-05-18
ApplicationBlueprinter pattern


toobulkeh:
Is there any reason not to use the `Application` pattern with Blueprinter?

`application_blueprint.rb`:
```ruby
class ApplicationBlueprint < Blueprinter::Base
  identifier :id
end
```

`user_blueprint.rb`:
```ruby
class UserBlueprint < ApplicationBlueprint
  fields :name
end
```

philipqnguyen:
Definitely! There is no reason not to use the `ApplicationBlueprint` pattern. It is already a pattern that Procore is using internally as well and I highly recommend using that pattern in your projects.

toobulkeh:
Apparently extending a blueprint with an identifier removes it from the returned value. Is that expected?

IE, I had to add this line back to all of our blueprints:

```ruby
identifier :id
```

philipqnguyen:
I did a short test of your example locally and I can confirm you're correct and that it does not work. Will  have to look into it further.

Systho:
The problem is the view collection is a class instance variable, thus it's not the same in the parent class and in the subclass.

I've personally fixed the problem with this snippet in my BaseBlueprint

```
  class BaseBlueprint < Blueprinter::Base

    def self.inherited(sub_class)
      sub_class.identifier :id
    end

end
```

philipqnguyen:
@Systho you're right, class instance variable is the problem. I think you have the right solution.

philipqnguyen:
resolved by #105

PR #105, opened 2018-10-13
Inherit view definition when using inheritance

hugopeixoto:
Enable field and view definitions on superclasses
to be used by subclasses:

    class ApplicationBlueprint < Blueprinter::Base
      identifier :id
      field :created_at
    end

    class PersonBlueprint < ApplicationBlueprint
      fields :name, :age
    end

Objects rendered with `PersonBlueprint` will contain fields `id`,
`created_at`, `name`, and `age`.

(Addresses issue #80)

### DIFF
--- a/lib/blueprinter/view.rb
+++ b/lib/blueprinter/view.rb
@@ -33,10 +33,6 @@ module Blueprinter
     end
 
     def <<(field)
-      if fields.has_key?(field.name)
-        raise BlueprinterError,
-          "Field #{field.name} already defined on #{name}"
-      end
       fields[field.name] = field
     end
   end
--- a/lib/blueprinter/view.rb
+++ b/lib/blueprinter/view.rb
@@ -10,6 +10,20 @@ module Blueprinter
       @excluded_field_names = excluded_view_names
     end
 
+    def inherit(view)
+      view.fields.values.each do |field|
+        self << field
+      end
+
+      view.included_view_names.each do |view_name|
+        include_view(view_name)
+      end
+
+      view.excluded_field_names.each do |field_name|
+        exclude_field(field_name)
+      end
+    end
+
     def include_view(view_name)
       included_view_names << view_name
     end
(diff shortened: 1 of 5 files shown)


hugopeixoto:
This may not work when reopening the superclass:

    class ApplicationBlueprint < Blueprinter::Base
      identifier :id
    end

    class PersonBlueprint < ApplicationBlueprint
      field :name 
    end

    class ApplicationBlueprint < Blueprinter::Base
      field :created_at 
    end

Rendering Person in this case would not contain `created_at`, since it was defined after inheriting. Not sure if we should address this usecase, but it would be a bit harder to implement. We'd have to resolve the view definition at rendering time, since the parent class may be redefined at any time.

philipqnguyen:
This is great work!

I don't think we need to solve for the use case of re-opening the super class, at least not now.

You can ignore #94, I'm still experimenting with that PR (I might close it for now actually so it doesn't cause confusion)

lib/blueprinter/view.rb
```diff
@@ -10,6 +10,20 @@ def initialize(name, fields: {}, included_view_names: [], excluded_view_names: [
       @excluded_field_names = excluded_view_names
     end
 
+    def inherit(view)
+      view.fields.values.each do |field|
+        self << field
```
mcclayton: I'll defer to @philipqnguyen for desired behavior, but due to the implementation of `<<` in this class, we'd raise an error if a field name collides, however, in inheritance -- you'd expect those fields to be overridable.

mcclayton: I agree, would be nice to have some specs to ensure that the last defined field is the field implementation that gets set. Both in a single view with multiple duplicate field name implementations, as well for the case where multiple classes define a field, to ensure the lowest subclass implementation gets set.

hugopeixoto: Nice catch. I'll add the specs and remove the guard, then.

hugopeixoto: I added a spec for the subclassing scenario and modified the existing spec for the regular overriding scenario. Let me know what you think

philipqnguyen:
We'll put this into 0.7.0

Issue #90, opened 2018-08-28
Pass a block to an association


Systho:
Sometimes I'd like to expose an association which is not directly  callable on the record itself. My main case is calling the association on a belongs_to or a has_one "middle" object.

I thought I could pass a block to the association like I do for the normal fields but it does not happen to work.

Do you think it would be possible for the `association`to work like the `field` and accept an optional block returning an AR::Relation ?

philipqnguyen:
Yes, I think this feature is needed. Would you (or anyone) be willing to make a PR? I would too, but it will take me a week or to before I can get to it, I've been a bit busy.

Systho:
Unfortunately I'm not sure of my availability during the next days, I may make a PR but it's very inprobable

philipqnguyen:
resolved by #106

PR #106, opened 2018-10-15
Allow associations to be defined with a block

hugopeixoto:
Addresses #90.

I removed an extra hash from `BlockExtractor`. Not sure if should have, but it seemed to be redundant, and tests are passing.

I changed `AssociationExtractor` so that it accepts a nested extractor. This allowed me to reuse `AutoExtractor` and `BlockExtractor`, making `#association` now very similar to `#field`.

I noticed that we're passing classes as extractors, and that the class method `extract` always creates a new instance of itself. This will probably create an unnecessary number of objects. Instantiating them directly in `#field` and `#association` might help improve performance. I addressed that in this PR as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,44 @@ Output:
 }
 ```
 
+#### Defining an association directly in the Blueprint
+
+You can also pass a block to an association:
+
+```ruby
+class ProjectBlueprint < Blueprinter::Base
+  identifier :uuid
+  field :name
+end
+
+class UserBlueprint < Blueprinter::Base
+  identifier :uuid
+
+  association :projects, blueprint: ProjectBlueprint do |user|
+    user.projects + user.company.projects
+  end
+end
+```
+
+Usage:
+
+```ruby
+puts UserBlueprint.render(user)
+```
+
+Output:
+
+```json
+{
+  "uuid": "733f0758-8f21-4719-875f-262c3ec743af",
+  "projects": [
+    {"uuid": "b426a1e6-ac41-45ab-bfef-970b9a0b4289", "name": "query-console"},
+    {"uuid": "5bd84d6c-4fd2-4e36-ae31-c137e39be542", "name": "blueprinter"},
+    {"uuid": "785f5cd4-7d8d-4779-a6dd-ec5eab440eff", "name": "uncontrollable"}
+  ]
+}
+```
+
 ### Passing additional properties to `render`
 
 `render` takes an options hash which you can pass additional properties, allowing you to utilize those additional properties in the `field` block. For example:
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -37,7 +37,7 @@ module Blueprinter
     #   end
     #
     # @return [Field] A Field object
-    def self.identifier(method, name: method, extractor: AutoExtractor)
+    def self.identifier(method, name: method, extractor: AutoExtractor.new)
       view_collection[:identifier] << Field.new(method, name, extractor, self)
     end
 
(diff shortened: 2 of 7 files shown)


lib/blueprinter/base.rb
```diff
@@ -129,9 +129,9 @@ def self.association(method, options = {}, &block)
       name = options.delete(:name) || method
 
       options = if block_given?
-        options.merge(extractor: BlockExtractor, block: block)
+        options.merge(extractor: BlockExtractor.new, block: block)
       else
-        options.merge(extractor: AutoExtractor)
+        options.merge(extractor: AutoExtractor.new)
```
philipqnguyen: Since you're already instantiating these `Extractor` classes, do you mind also instantiating the `Extractor` in `fields` and `identifier` 
here https://github.com/procore/blueprinter/blob/master/lib/blueprinter/base.rb#L41
and here https://github.com/procore/blueprinter/blob/master/lib/blueprinter/base.rb#L219
and then remove this https://github.com/procore/blueprinter/blob/master/lib/blueprinter/extractor.rb#L11-L13

lib/blueprinter/base.rb
```diff
@@ -124,14 +124,21 @@ def self.field(method, options = {}, &block)
...
+        options.merge(extractor: AutoExtractor)
+      end
+
       current_view << Field.new(method,
-                                       name,
-                                       AssociationExtractor,
-                                       self,
-                                       options.merge(association: true))
+                                name,
+                                AssociationExtractor.new(options[:extractor]),
```
philipqnguyen: I don't think you need to pass the `options[:extractor]` here, the `AssociationExtractor` should already be passed the `options` from `Field` https://github.com/procore/blueprinter/blame/master/lib/blueprinter/field.rb#L13

lib/blueprinter/extractors/association_extractor.rb
```diff
@@ -1,8 +1,12 @@
 module Blueprinter
   class AssociationExtractor < Extractor
+    def initialize(field_extractor)
+      @field_extractor = field_extractor
+    end
+
     def extract(association_name, object, local_options, options={})
-      value = object.public_send(association_name)
-      return (value || options[:default]) if value.nil?
+      value = @field_extractor.extract(association_name, object, local_options, options)
```
philipqnguyen: you should be able to do `options[:extractor].new.extract( .... )`

philipqnguyen:
Would you also add this feature to the README and add the method documentation above `def self.association`? Similar to how it's documented for `field` blocks here https://github.com/procore/blueprinter/blob/master/lib/blueprinter/base.rb#L67-L80

lib/blueprinter/base.rb
hugopeixoto: Removing https://github.com/procore/blueprinter/blob/master/lib/blueprinter/extractor.rb#L11-L13
makes this break: https://github.com/procore/blueprinter/blob/master/spec/integrations/shared/base_render_examples.rb#L41

This may be a breaking change, so I'm not sure if we want to change this here. Maybe I should just apply the other changes you mentioned but leave `self.extract`?

philipqnguyen: I didn't realize there were tests for that. OK let's leave that for now. Thanks!

README.md
```diff
@@ -186,6 +186,44 @@ Output:
...
+puts UserBlueprint.render(user)
+```
+
+Output:
+
+```json
+{
+  "uuid": "733f0758-8f21-4719-875f-262c3ec743af",
+  "projects": [
+    {"uuid": "b426a1e6-ac41-45ab-bfef-970b9a0b4289", "name": "query-console"},
```
mcclayton: I see you've seen my library 😂

philipqnguyen:
released with 0.7.0

PR #107, opened 2018-10-17
Use field in fields

hugopeixoto:
This allows us to make changes in a single place.

It also allows overriding #field without having to worry about overriding #fields


philipqnguyen:
looks like conflict from the #106 merge

hugopeixoto:
Yeah, I was expecting that. Fixed it :)

PR #108, opened 2018-10-18
Update docs regarding the args yielded to blocks

philipqnguyen:
The blocks are yielded both the object and the options that were passed
to render. This commit simply ensure that the docs reflects that.

### DIFF
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -68,8 +68,8 @@ module Blueprinter
     #   to call to determine if the field should be included (e.g.
     #   `unless: :include_first_name?, or unless: Proc.new { |user, options| options[:current_user] != user }).
     #   The method, proc or string should return or evaluate to a true or false value.
-    # @yield [Object] The object passed to `render` is also passed to the
-    #   block.
+    # @yield [object, options] The object and the options passed to render are
+    #   also yielded to the block.
     #
     # @example Specifying a user's first_name to be serialized.
     #   class UserBlueprint < Blueprinter::Base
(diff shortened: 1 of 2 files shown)


mcclayton:
Very nice 👍

lib/blueprinter/base.rb
```diff
@@ -119,8 +121,8 @@ def self.field(method, options = {}, &block)
     #   JSON output.
     # @option options [Symbol] :view Specify the view to use or fall back to
     #   to the :default view.
-    # @yield [Object] The object passed to `render` is also passed to the
-    #   block.
+    # @yield [object, options] The object and the options passed to render is
```
mcclayton: Oh, perhaps `are` instead of `is`.

lib/blueprinter/base.rb
```diff
@@ -68,8 +68,8 @@ def self.inherited(subclass)
     #   to call to determine if the field should be included (e.g.
     #   `unless: :include_first_name?, or unless: Proc.new { |user, options| options[:current_user] != user }).
     #   The method, proc or string should return or evaluate to a true or false value.
-    # @yield [Object] The object passed to `render` is also passed to the
-    #   block.
+    # @yield [object, options] The object and the options passed to render is
```
mcclayton: Oh, perhaps `are` instead of `is`.

philipqnguyen: ty

lib/blueprinter/base.rb
philipqnguyen: ty

Issue #109, opened 2018-10-19
Should `identifier` accept a block?


hugopeixoto:
Similarly to how `field` and `association` now accept a block, should `identifier` also be able to accept a block?

mcclayton:
I like the idea and I think this certainly will provide a greater deal of flexibility while also making the API more consistent so I am a +1.

mcclayton:
Resolved by: https://github.com/procore/blueprinter/pull/110

PR #110, opened 2018-10-20
Allow identifiers to be defined with a block

hugopeixoto:
Now that `::association` and `::field` receive an optional block,
`::identifier` should receive it as well.

Since all three methods receive an optional block, the block extraction
mechanism was moved to `AutoExtractor`. Removing this mechanism from
each individual method made them very similar. `::association` now uses
`::field` instead of directly adding the field to the current view.

Addresses #109

### DIFF
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -26,6 +26,9 @@ module Blueprinter
     # @param name [Symbol] to rename the identifier key in the JSON
     #   output. Defaults to method given.
     # @param extractor [AssociationExtractor,AutoExtractor,BlockExtractor,HashExtractor,PublicSendExtractor]
+    # @yield [object, options] The object and the options passed to render are
+    #   also yielded to the block.
+    #
     #   Kind of extractor to use.
     #   Either define your own or use Blueprinter's premade extractors.
     #   Defaults to AutoExtractor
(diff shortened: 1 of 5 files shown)


lib/blueprinter/base.rb
```diff
@@ -37,12 +37,14 @@ class Base
     #   end
     #
     # @return [Field] A Field object
-    def self.identifier(method, name: method, extractor: AutoExtractor.new)
-      view_collection[:identifier] << Field.new(method, name, extractor, self)
-    end
-
-    def self.inherited(subclass)
-      subclass.send(:view_collection).inherit(view_collection)
+    def self.identifier(method, name: method, extractor: AutoExtractor.new, &block)
```
mcclayton: Would you mind adding to the method documentation for this and adding it to the README similar to how you did for the associations?
https://github.com/procore/blueprinter/blob/f849049fedf7adea5f13d11b2514b6c8b54f7efa/lib/blueprinter/base.rb#L124-L139
https://github.com/procore/blueprinter/blob/master/README.md#defining-an-association-directly-in-the-blueprint

lib/blueprinter/base.rb
```diff
@@ -313,6 +305,11 @@ def self.view(view_name)
 
     private
 
+    def self.inherited(subclass)
+      subclass.send(:view_collection).inherit(view_collection)
+    end
+    private_class_method :inherited
```
AllPurposeName: Great catch!

hugopeixoto:
I've added documentation. Feel free to suggest better examples, I used the one I needed on my project, but it may be a bit weird.

philipqnguyen:
LGTM!

mcclayton:
We’ll hopefully get this out soon in the next release of Blueprinter, thanks for this!

PR #112, opened 2018-10-29
Base.associations should be completely private

philipqnguyen:
Base.associations is not being used outside of Base, and should
not be exposed.


philipqnguyen:
@wai-procore @AllPurposeName @dlcarter @adamwells want to give your code review? This small change so =D

PR #115, opened 2018-11-13
Accept an option 'default' to allow setting default serialized field value if null

mcclayton:
**Included in the PR**

For fields that have a value of nil, blueprinter serializes to `null` by default. This PR adds an ability to add a `default` option to blueprinter fields which can be used as the serialized value instead of null.

**Before**
```ruby
  identifier :uuid
  field :first_name
end
```

`UserBlueprint.render(user)` renders `{ "id": 1, "first_name": null }`

**After**
```ruby
  identifier :uuid
  field :first_name, default: "N/A"
end
```

`UserBlueprint.render(user)` renders `{ "id": 1, "first_name": "N/A" }`

Aims to address: https://github.com/procore/blueprinter/issues/114

### DIFF
--- a/lib/blueprinter/extractors/association_extractor.rb
+++ b/lib/blueprinter/extractors/association_extractor.rb
@@ -5,15 +5,10 @@ module Blueprinter
     end
 
     def extract(association_name, object, local_options, options={})
-      if options.key?(:default)
-        value = @extractor.extract(association_name, object, local_options, options.except(:default))
-        return options[:default] if value.nil?
-      else
-        value = @extractor.extract(association_name, object, local_options, options)
-        return options[:default] if value.nil?
-        view = options[:view] || :default
-        options[:blueprint].prepare(value, view_name: view, local_options: local_options)
-      end
+      value = @extractor.extract(association_name, object, local_options, options.except(:default))
+      return options[:default] if value.nil?
+      view = options[:view] || :default
+      options[:blueprint].prepare(value, view_name: view, local_options: local_options)
     end
   end
 end
--- a/lib/blueprinter/extractors/association_extractor.rb
+++ b/lib/blueprinter/extractors/association_extractor.rb
@@ -5,10 +5,16 @@ module Blueprinter
     end
 
     def extract(association_name, object, local_options, options={})
-      value = @extractor.extract(association_name, object, local_options, options)
-      return options[:default] if value.nil?
-      view = options[:view] || :default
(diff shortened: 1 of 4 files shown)


lib/blueprinter/extractors/association_extractor.rb
```diff
@@ -5,7 +5,7 @@ def initialize
     end
 
     def extract(association_name, object, local_options, options={})
-      value = @extractor.extract(association_name, object, local_options, options)
+      value = @extractor.extract(association_name, object, local_options, options.except(:default))
```
philipqnguyen: Why do you need to do `except(:default)`?

lib/blueprinter/extractors/association_extractor.rb
```diff
@@ -5,7 +5,7 @@ def initialize
...
       return options[:default] if value.nil?
```
philipqnguyen: Can we just let the `AutoExtractor` deal with the options[:default] stuff? And take it out of here?

lib/blueprinter/extractors/association_extractor.rb
mcclayton: I need to `except(:default)` here, because otherwise, when it makes the call through to the AutoExtractor, the AutoExtractor (with these changes) would then return the default value instead of `nil`. Then the association extractor would not pass the `return options[:default] if value.nil?` step and it would go on to call `options[:blueprint].prepare(value, view_name: view, local_options: local_options)`. You can verify that this would fail the test for default association values.

Another option (which is closer to what you are describing) we be something like this:
```
def extract(association_name, object, local_options, options={})
      value = @extractor.extract(association_name, object, local_options, options)
      if (options.key?(:default) || value.nil?)
        return value
      else
        options[:blueprint].prepare(value, view_name: view, local_options: local_options)
      end
    end
```

Thoughts?

lib/blueprinter/extractors/association_extractor.rb
mcclayton: Yeah I proposed an alternative solution above that let's the AutoExtractor do all the work necessary for defaults.

lib/blueprinter/extractors/association_extractor.rb
philipqnguyen: OHH I see ok nevermind. I think what you have is fine.

lib/blueprinter/extractors/association_extractor.rb
philipqnguyen: nevermind

PR #117, opened 2018-11-14
do not raise error on null date with date_format option (closes #116)

### LINKED ISSUE #116: Rendering null date with date_format option causes error
Hi, thanks for the gem! We're choosing library to replace roar and we've faced with the next issue:

```ruby
class A < Blueprinter::Base
  field :date, datetime_format: '%FT%T%:z'
end

# this causes Blueprinter::BlueprinterError Cannot format invalid DateTime object
A.render_as_hash(OpenStruct.new(date: nil))
```

I suppose it should render nil as is. I'll open PR soon :)

tpltn:


spec/integrations/base_spec.rb
```diff
@@ -186,6 +186,22 @@
...
+    end
+
+    context 'field with nil instead of value' do
+      let(:object) { OpenStruct.new(date: nil) }
+
+      it 'renders null' do
+        expect(blueprint.render_as_hash(object)).to eq(date: nil)
+      end
+    end
+  end
```
philipqnguyen: There's already a context for datetime. Would you augment that context to include this test instead of putting this test here? 

https://github.com/procore/blueprinter/blame/master/spec/integrations/shared/base_render_examples.rb#L53-L64

tpltn: Done)

PR #118, opened 2018-11-15
Extend Support for other JSON encoders like yajl-ruby

ritikesh:


lib/blueprinter/configuration.rb
```diff
@@ -1,9 +1,14 @@
...
-    attr_accessor :generator
+    attr_accessor :generator, :method
 
     def initialize
       @generator = JSON
+      @method = :generate
+    end
+
+    def jsonify(blob)
+      generator.send(method, blob)
```
philipqnguyen: Thank you for doing this! Just one minor request. Can we use `public_send` over `send`?

ritikesh: Thanks for getting back on this. I've changed it to use public_send. Let me know if you need anything else.

ritikesh:
@philipqnguyen Thanks for approving the PR. Any updates on when this will be merged?

philipqnguyen:
I don't do the releases anymore, but @mcclayton does!

mcclayton:
Thank you so much for the contribution @ritikesh! These changes look good to me and I'll merge these now, I will try to draft the next blueprinter release later today 👍

PR #119, opened 2018-11-21
render_as_json, making sorting fields optional

ritikesh:
1. Added a `render_as_json` API - When you want a JSONified hash - useful when you want to modify the generated hash but require it to be JSONified 
2. Added a configuration option to make sorting optional - for those who don't want to sort the fields by name but want it in the order it was defined. Bonus - adds a perf benefit. 
3. Update readme with latest updates and some missing public API specs like `render_as_hash`, `exclude`..

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,13 +399,13 @@ $ gem install blueprinter
 You should also have `require 'json'` already in your project if you are not using Rails or if you are not using Oj.
 
 ## Sorting
-By default the response sorts the keys by name. If you want the fields to be sorted in the order of definition, set the below configuration to true.
+By default the response sorts the keys by name. If you want the fields to be sorted in the order of definition, use the below configuration option.
 
 Usage:
 
 ```ruby
 Blueprinter.configure do |config|
-  config.sort_by_definition = true
+  config.sort_fields_by = :definition
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -278,6 +278,42 @@ Output:
 }
 ```
 
+### render_as_hash
+Same as `render`, returns a Ruby Hash.
+
+Usage:
+
+```ruby
+puts UserBlueprint.render_as_hash(user, company: company)
+```
+
+Output:
+
+```ruby
+{
+  uuid: "733f0758-8f21-4719-875f-262c3ec743af",
+  company_name: "My Company LLC"
+}
+```
+
+### render_as_json
+Same as `render`, returns a Ruby Hash JSONified. This will call JSONify all keys and values.
+
+Usage:
+
+```ruby
+puts UserBlueprint.render_as_json(user, company: company)
+```
+
+Output:
+
+```ruby
+{
+  "uuid" => "733f0758-8f21-4719-875f-262c3ec743af",
+  "company_name" => "My Company LLC"
+}
+```
+
 ### Conditional field
 
 `field` supports `:if` and `:unless` options argument that can be used to serialize the field conditionally.
@@ -328,6 +364,34 @@ $ gem install blueprinter
 
 You should also have `require 'json'` already in your project if you are not using Rails or if you are not using Oj.
 
+## Sorting
+By default the response sorts the keys by name. If you want the fields to be sorted in the order of definition, set the below configuration to true.
+
+Usage:
+
+```ruby
+Blueprinter.configure do |config|
+  config.sort_by_definition = true
+end
+```
+
+```ruby
+class UserBlueprint < Blueprinter::Base
+  identifier :name
+  field :email
+  field :birthday, datetime_format: "%m/%d/%Y"
+end
+```
+
+Output:
+```json
+{
+  "name": "John Doe",
+  "email": "john.doe@some.fake.email.domain",
+  "birthday": "03/04/1994"
+}
+```
+
 ## OJ
 
 By default, Blueprinter will be calling `JSON.generate(object)` internally and it expects that you have `require 'json'` already in your project's code. You may use `Oj` to generate in place of `JSON` like so:
@@ -352,7 +416,7 @@ gem 'oj'
 [yajl-ruby](https://github.com/brianmario/yajl-ruby) is a fast and powerful JSON generator/parser. To use `yajl-ruby` in place of `JSON / OJ`, use:
 
 ```ruby
-require 'yajl' # you can skip this if OJ has already been required.
+require 'yajl' # you can skip this if yajl has already been required.
 
 Blueprinter.configure do |config|
   config.generator = Yajl::Encoder # default is JSON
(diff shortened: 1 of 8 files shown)


ritikesh:
@philipqnguyen @mcclayton Hi guys, any updates here?

ritikesh:
@philipqnguyen @mcclayton 
Hi guys, any updates here?

README.md
```diff
@@ -328,6 +398,34 @@ $ gem install blueprinter
...
 You should also have `require 'json'` already in your project if you are not using Rails or if you are not using Oj.
 
+## Sorting
+By default the response sorts the keys by name. If you want the fields to be sorted in the order of definition, set the below configuration to true.
+
+Usage:
+
+```ruby
+Blueprinter.configure do |config|
+  config.sort_by_definition = true
```
mcclayton: This is probably fine as I don't think we need any more types of sorting, but this may be a little more future-proof/extensible and a little bit better named if it were:
`config.sort_fields_by` where the options could be `[:name_asc, :definition]` with `:name_asc` being the default.

ritikesh:
> Thanks so much for your valuable contribution @ritikesh. Apologies for the delay in reviewing, been a little slow with the holidays 🦃

No worries, hope had a good time there. 😄

lib/blueprinter/configuration.rb
```diff
@@ -1,10 +1,11 @@
 module Blueprinter
   class Configuration
-    attr_accessor :generator, :method
+    attr_accessor :generator, :method, :sort_by_definition
 
     def initialize
       @generator = JSON
       @method = :generate
+      @sort_by_definition = false
```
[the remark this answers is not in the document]
ritikesh: just figured a boolean comparison would be easier and lighter for execution [here](https://github.com/procore/blueprinter/pull/119/files#diff-d909c0537c86c6300c32c53da4d2c8c5R58). Going by that context, I couldn't think of a more apt variable name. If you feel strongly otherwise, I could go ahead with your approach as well.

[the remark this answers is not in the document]
philipqnguyen: I like @mcclayton suggestion too!

lib/blueprinter/version.rb
```diff
@@ -1,3 +1,3 @@
 module Blueprinter
-  VERSION = '0.8.0'
+  VERSION = '0.9.0'
```
philipqnguyen: Would you also update the CHANGELOG.md since you're already bumping the version here? =)

ritikesh: will do.

ritikesh:
@mcclayton I've made the changes as suggested. Please let me know if you need anything else.

README.md
```diff
@@ -328,6 +398,34 @@ $ gem install blueprinter
 
 You should also have `require 'json'` already in your project if you are not using Rails or if you are not using Oj.
 
+## Sorting
+By default the response sorts the keys by name. If you want the fields to be sorted in the order of definition, use the below configuration option.
+
```
mcclayton: If you could just document the options available here, I think that's all I'd request and looks good to me. Something like:
`sort_fields_by ` options:
  - `:name_asc` (default): Sorts the fields in ascending order by field name.
  - `:definition`: Preserves the order the fields were defined in.

ritikesh: :name_asc is default. That's why I skipped it and gave an example for :definition alone.

lib/blueprinter/view_collection.rb
```diff
@@ -20,7 +21,11 @@ def has_view?(view_name)
     end
 
     def fields_for(view_name)
-      identifier_fields + sortable_fields(view_name).values.sort_by(&:name)
+      sorted_fields = sortable_fields(view_name).values
```
AllPurposeName: This would both be more clear to me and leave ourselves open to change if we assign the variable only once (conditionally) before adding it to `identifier_fields`

ie
```
sorted_fields = if sort_by_definition
  sortable_fields(view_name).values.sort_by(&:name)
else
  sortable_fields(view_name).values
end
identifier_fields + sorted_fields
```

What do you think?

lib/blueprinter/view_collection.rb
```diff
@@ -48,5 +53,13 @@ def sortable_fields(view_name)
 
       fields
     end
+
+    def merge_fields(source_fields, included_fields)
+      unless sort_by_definition
```
AllPurposeName: Can we check for the positive case rather than using `unless`?

AllPurposeName: Neither?

```
if sort_by_definition
  included_fields.merge(source_fields)
else
  source_fields.merge(included_fields)
end
```

lib/blueprinter/view_collection.rb
AllPurposeName: Alright I can agree to set the `sorted_fields` variable at the beginning of the method, but I still don't think we should reassign the variable within a condition. If we want to conduct an operation on this collection conditionally, let's be honest about it.

```
sorted_fields = sortable_fields(view_name).values

sorted_fields.sort_by!(:name) unless sort_by_definition

identifier_fields + sorted_fields
```

Alternatively we could treat these two variables as separate things entirely:
```
fields_sorted_by_definition = sortable_fields(view_name).values

fields_sorted_by_name = fields_sorted_by_definition.sort_by(&:name) unless sort_by_definition

identifier_fields + (fields_sorted_by_name || fields_sorted_by_definition)
```

I think the second option is cumbersome, but I would rather have to read more than do a double take at a variable being reassigned in a condition.

lib/blueprinter/view_collection.rb
ritikesh: Sorry, I was browsing this on my mobile and carried the discussion from our previous conversation to this in my reply. 🙈 Looks straightforward, I'll change that. Do you have any other inputs/questions?

lib/blueprinter/view_collection.rb
ritikesh: I agree with your philosophy from the first suggestion and that was my goto option as well, but there ins't a `sort_by!` method available. 😞Hence the current implementation.
From the second suggestion, feels like we're adding one too many variables. Why not reuse the same one? - similar to approach one where we're modifying the original variable itself. They're all bound to the same method anyway and have no usage outside the scope of this method.

philipqnguyen:
``` ruby
fields = sortable_fields(view_name).values
sorted_fields = sort_by_definition ? fields.sort_by(:name) : fields
identifier_fields + sorted_fields
```
?

ritikesh: Okay, I like your approach and I can make these changes. 

But it's still creating another local variable. Is reassigning to the same local variable considered bad practice? I tend to do that often to avoid having too many variables and think which one refers to what since at the end I'm only bothered about what the method returns.

philipqnguyen: I try to make new variables if possible rather reassigning them. Reassigning just makes it harder to follow in my opinion.

PR #122, opened 2018-12-20
Dynamically evaluate Blueprints for associations

ritikesh:
Useful when we have polymorphic associations and want to reuse existing blueprints.


ritikesh:
@philipqnguyen @mcclayton 
Hi guys, hope you had a good Christmas. Just wanted to follow up on this. 
Again, thanks your patience and response on the previous PRs. Much appreciated.

README.md
```diff
@@ -192,6 +192,20 @@ class UserBlueprint < Blueprinter::Base
 end
 ```
 
+### Supporting Dynamic Blueprints for associations
+When defining a association in the blueprint, we can dynamically evaluate the blueprint as well. This comes in handy when adding `polymorphic associations` to the blueprint but wanting to reuse existing blueprints.
```
njbbaer: Just some suggested language changes,
```
When defining an association, we can dynamically evaluate the blueprint. This comes in handy when adding polymorphic associations, by reusing the existing blueprints.
```

ritikesh:
> Hi @ritikesh, thanks for contributing again.
> 
> I think I understand what this PR is doing, but I'm not convinced of its usefulness. Can you provide a more detailed or real world example here than what you put in the readme?

Hi @njbbaer,

Sure, it's fairly common for real world rails apps to have polymorphic associations defined in models. And one would also want to export that data in the payload in a dynamic and readable format. For eg, consider I have a `tasks` model which is `polymorphic` and can belong to a `ticket`, `project`, etc. 

In this sense, if I have a `Task blueprint` and want to depict meaningful information(which could be different for each kind of `taskable`) about the associated `taskable` object, I would want to reuse my existing blueprints and not duplicate things.
```ruby
class TicketBlueprint < Blueprinter::Base
	view :base_skeleton do
		fields :id, :subject, :created_at, :updated_at
	end
	view :polymorphic do
		include_view :base_skeleton
		field :priority, :assigned_agent, :status
	end
end
class ProjectBlueprint < Blueprinter::Base
	view :polymorphic do
		fields :id, :title, :due_by, :progress, :created_at, :updated_at
		association :project_head, blueprint: UserBlueprint
	end
end
class TaskBlueprint < Blueprinter::Base
	...
	# taskable.blueprint -> TicketBlueprint / ProjectBlueprint
	association :taskable, blueprint: ->(taskable) { taskable.blueprint }, view: :polymorphic, default: {}
end
```

njbbaer:
Thanks for providing that example.

The only issue I see with this is that it probably isn't good practice for an API, since the shape of `taskable` can change which may not be anticipated and is tricky to document.

But since Blueprinter is a generic JSON serializer (not just for APIs) I think this feature still makes sense and we should include it.

ritikesh:
Our primary usage of blueprint at the moment is to generate JSON payloads for multiple downstream microservices which requires us to push different kinds of metadata for the taskable association. 

We are planning to extend it to our public API JSON generators as well in the future, in which case, we'd probably be using different views altogether and stick to a standard structure.

lib/blueprinter/version.rb
```diff
@@ -1,3 +1,3 @@
 module Blueprinter
-  VERSION = '0.9.0'
+  VERSION = '0.9.1'
```
philipqnguyen: Should this be a 0.10.0 since it's new feature?

README.md
```diff
@@ -192,6 +191,20 @@ class UserBlueprint < Blueprinter::Base
...
+  identifier :uuid
+
+  view :normal do
+    field :title, default: "N/A"
+    association :taskable, blueprint: ->(association) {association.blueprint}, default: {}
+  end
+end
+```
+Note: `association.blueprint` should return a valid Blueprint class. Currently, `has_many` is not supported because of the very nature of polymorphic associations.
+
```
philipqnguyen: Can you also add an example on how to define `blueprint` on the associated model?

.gitignore
```diff
@@ -8,3 +8,4 @@ spec/dummy/tmp/
 doc/
 .yardoc/
 Gemfile.lock
+.vscode
```
philipqnguyen: I think you need a newline here

spec/integrations/base_spec.rb
```diff
@@ -85,6 +85,19 @@
...
         end
+        context "Given association with dynamic blueprint" do
+          let(:blueprint) do
+            Class.new(Blueprinter::Base) do
+              association :user, blueprint: ->(obj) { user_blueprint }
+              def self.user_blueprint
+                Class.new(Blueprinter::Base) { identifier :id }
+              end
+            end
+          end
```
philipqnguyen: Would you redo the test so that it follows more closely what the use case should look like? For example: 
`association :user, blueprint: ->(obj) { obj.blueprint }`

philipqnguyen:
Also sorry for coming in late to this PR... was on a long and mostly computer-less break.

spec/integrations/base_spec.rb
ritikesh: I was having a very hard time making the user/vehicle object respond to a blueprint method. Since the feature was to let association blueprint be fetched dynamically, I took this route instead.

lib/blueprinter/version.rb
ritikesh: Thought simpler features could be accommodated in patch version updates itself.

spec/integrations/base_spec.rb
philipqnguyen: You should be able to reopen the User class and add `blueprint` to it.

Try this:
```ruby
        context "Given association with dynamic blueprint" do
            fields :id
          end
          class User < ActiveRecord::Base
            def blueprint
              UserBlueprint
            end
          end
          let(:blueprint) do
              association :user, blueprint: ->(obj) { obj.blueprint }
            end
          end
          it "should render the association with dynamic blueprint" do
            expect(JSON.parse(blueprint.render(vehicle))["user"]).to eq({"id"=>obj.id})
          end
        end
```

ritikesh: I tried various ways, `patching`, `class_eval` etc. Nothing helped, so gave up. I'll give this another shot today! 😄

ritikesh:
Hi guys, is there anything else pending on this?

mcclayton:
Thanks @ritikesh! I'll merge this now but probably won't be able to release Blueprinter until next week due to some organizational processes.

ritikesh:
No worries! Thanks for merging, @mcclayton. 👍

gogvale:
It would be awesome if someone could add this example to the README

Issue #124, opened 2018-12-27
skip items with nil values


skcc321:
Is it possible to have an option for ignoring nil values like this one in JBuilder https://www.rubydoc.info/github/rails/jbuilder/Jbuilder:ignore_nil!
I'm using JBuilder everywhere, and we have a convention to skip items with nil values (client wants it). I'm going to use Blueprinter instead, but the last thing which blocks to do that is this missing option.

mcclayton:
Hi @skcc321, thanks for reaching out. 
I'll bring this up to the team and see if (and in what way if so) we'd like to support something like this.
In the meantime, this is something that can currently be accomplished using Blueprinter's [conditional field functionality](https://github.com/procore/blueprinter#conditional-field).
i.e.: 

```
field :name, unless: ->(user, _options) { user.name.nil? }
```

Though obviously this is not ideal to have to specify for every single one of your Blueprint fields. I'll re-post when I get the team's feedback on this.

anlek:
I have had this issue with `date_format: "%d %m %Y"` and the date being `nil`.  It would be great if the result would be `null` instead of an error about not being able to convert `nil` to a date object.

mcclayton:
Hi @anlek, can you please verify for me what version of Blueprinter you are using? That specific issue you are seeing for nil dates ought to be fixed in versions >= `0.8.0`.
https://github.com/procore/blueprinter/pull/117
https://github.com/procore/blueprinter/blob/master/CHANGELOG.md#080----20181119

mcclayton:
@skcc321 
Sorry for the delay in getting back to you, I brought it up to the team and we are considering a more general solution that would hopefully suffice your issue.
We are considering implementing global `:if` and `:unless` options in the Blueprinter config.
i.e.
```ruby
config.if = -> { obj.is_a?(Foo) }
```

```ruby
class SomeBluepint
  field :stuff, if: -> { obj.is_a?(Bar) }
end
```
Where field-level configurations will override the global configuration, but the global configuration would be used if no field-level option was provided. 
I think implementing this would solve your issue if I am understanding you correctly?

anlek:
@mcclayton You're right, I was using `0.6.0` version. I'll update and look forward to the fix. As for the global if/unless. It's an interesting idea, I think it's a good place to build a catch all for a given application. Thanks for the details!

skcc321:
Thank you @mcclayton for taking care of this issue. I appreciate it a lot.
[...]

PR #127, opened 2019-01-15
Global if/unless Procs for Conditional Field Rendering

mcclayton:
Currently, you can conditionally render fields by setting a field-level options argument `:if` or `:unless`.
This PR enables the setting of a global `:if` and `:unless` proc that will be used to evaluate the conditional render of all fields.
The field-level setting overrides the global config setting (for the field) if both are set.

#### Field-level Setting
```ruby
  identifier :uuid
  field :last_name, if: ->(user, _field_name, options) { user.first_name != options[:first_name] }
  field :age, unless: ->(user, _field_name, _options) { user.age < 18 }
end
```

#### Global Config Setting
```ruby
  config.if = ->(obj, field_name, _options) { !obj[field_name].nil? }
  config.unless = ->(obj, field_name, _options) { obj[field_name].nil? }
end

```

This PR also addresses making https://github.com/procore/blueprinter/issues/124 easier/cleaner to accomplish

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,8 +381,8 @@ Both the `field` and the global Blueprinter Configuration supports `:if` and `:u
 #### Global Config Setting
 ```ruby
 Blueprinter.configure do |config|
-  config.if = ->(obj, field_name, _options) { !obj[field_name].nil? }
-  config.unless = ->(obj, field_name, _options) { obj[field_name].nil? }
+  config.if = ->(obj, _options) { obj.is_a?(Foo) }
+  config.unless = ->(obj, _options) { obj.is_a?(Bar) }
 end
 ```
 
@@ -390,8 +390,8 @@ end
 ```ruby
 class UserBlueprint < Blueprinter::Base
   identifier :uuid
-  field :last_name, if: ->(user, _field_name, options) { user.first_name != options[:first_name] }
-  field :age, unless: ->(user, _field_name, _options) { user.age < 18 }
+  field :last_name, if: ->(user, options) { user.first_name != options[:first_name] }
+  field :age, unless: ->(user, _options) { user.age < 18 }
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ end
 
 class Project < ActiveRecord::Base
   has_many :tasks, as: :taskable
-  
+
   def blueprint
     ProjectBlueprint
   end
@@ -374,10 +374,19 @@ Output:
 }
 ```
 
-### Conditional field
+### Conditional fields
 
-`field` supports `:if` and `:unless` options argument that can be used to serialize the field conditionally.
+Both the `field` and the global Blueprinter Configuration supports `:if` and `:unless` options that can be used to serialize fields conditionally.
 
+#### Global Config Setting
+```ruby
+Blueprinter.configure do |config|
+  config.if = ->(obj, _options) { obj.is_a?(Foo) }
+  config.unless = ->(obj, _options) { obj.is_a?(Bar) }
+end
+```
+
+#### Field-level Setting
 ```ruby
 class UserBlueprint < Blueprinter::Base
   identifier :uuid
@@ -385,6 +394,8 @@ class UserBlueprint < Blueprinter::Base
   field :age, unless: ->(user, _options) { user.age < 18 }
 end
 ```
+
+The field-level setting overrides the global config setting (for the field) if both are set.
 
 ### Custom formatting for dates and times
 To define a custom format for a Date or DateTime field, include the option `datetime_format` with the associated `strptime` format.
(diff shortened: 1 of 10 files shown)


lib/blueprinter/configuration.rb
```diff
@@ -1,16 +1,28 @@
...
       generator.public_send(method, blob)
     end
+
+    def valid_callable?(callable_name)
+      VALID_CALLABLES.include?(callable_name)
+    end
+  end
+
+  def self.reset_configuration!
+    @configuration = Configuration.new
```
philipqnguyen: What's the use for this? Is it only to clear the config for the tests?

mcclayton: I'm glad you pointed this out -- meant to leave a comment. That was the main primary motivating factor was that I wanted a more consistent way to reset back to the defaults. I re-wrote this to use a SpecHelper that does not modify the original class. Thanks for pointing out this code smell 👃

mcclayton:
@philipqnguyen Yes, I called this out as breaking in the CHANGELOG, The need is that, since the global `:if` and `:unless` procs are called at field-evaluation time, there's no way to see what field is being evaluated in the proc. So there'd be no way to accomplish #124 without passing in the field_name to do:
```
  config.if = ->(obj, field_name, _options) { !obj[field_name].nil? }
  config.unless = ->(obj, field_name, _options) { obj[field_name].nil? }
end
```

Make sense? Or am I missing a way to still be able to do this without passing in the field_name?

mcclayton:
Yeah we can totally merge the field_name into the option if you think that's kosher. I think that's a little unexpected to a user that we'd add it to the options as options are something that the *user* should be defining not that we'd ad-hoc add to. I definitely get your desire to avoid this sort of breaking change though. Global if/unless procs are not fully useful without the field_name though 🤔

philipqnguyen:
do you think we may have jumped the gun with globalizing `:if`/`:unless`. 

Maybe all we needed as a solution for #124 was an `:ignore_nil` option? It can be on both the field level and on the configuration level?

AllPurposeName:
I don't think we've jumped the gun, nor do I like the `field_name` in `options` workaround. This solution adds bloat-free useful functionality. Doing that workaround would be a little bit messy and definitely a little bloaty. Also, overloading options is always a slippery slope.

On a positive note, I have full confidence that we could move to 1.0.0 anytime and be fine adhering to semver with a bit more caution/maturity. Alternatively, you're right, it's not a 1.0 yet so we can make changes like this. If anyone needs to, they can pin their version to 0.10.0

README.md
```diff
@@ -374,18 +374,29 @@ Output:
...
 class UserBlueprint < Blueprinter::Base
   identifier :uuid
-  field :last_name, if: ->(user, options) { user.first_name != options[:first_name] }
-  field :age, unless: ->(user, _options) { user.age < 18 }
+  field :last_name, if: ->(user, _field_name, options) { user.first_name != options[:first_name] }
+  field :age, unless: ->(user, _field_name, _options) { user.age < 18 }
 end
 ```
 
+The field-level setting overrides the global config setting (for the field) if both are set.
```
AllPurposeName: Not blocking: This mention might be worth a **bolding**.

lib/blueprinter/configuration.rb
```diff
@@ -1,16 +1,24 @@
 module Blueprinter
   class Configuration
-    attr_accessor :generator, :method, :sort_fields_by
+    attr_accessor :generator, :if, :method, :sort_fields_by, :unless
+
+    VALID_CALLABLES = %i(if unless)
```
AllPurposeName: Can we `.freeze` this array?
```suggestion
    VALID_CALLABLES = %i(if unless).freeze
```

lib/blueprinter/field.rb
```diff
@@ -31,9 +31,17 @@ def unless_callable
   end
 
   def callable_from(option_name)
-    return false unless options.key?(option_name)
+    config = Blueprinter.configuration
+
+    # Use field-level callable, or when not defined, try global callable
+    tmp = if options.key?(option_name)
```
AllPurposeName: Non blocking: I know you didn't make it, but I don't like this variable name. since it's only invoked with `:if` or `:unless` I would be for renaming it to `condition` or `additional_condition`.

lib/blueprinter/version.rb
```diff
@@ -1,3 +1,3 @@
 module Blueprinter
-  VERSION = '0.10.0'
+  VERSION = '0.11.0'
```
AllPurposeName: Pending the conversation this may become `1.0.0`

philipqnguyen:
Ok, I'm good with this becoming 1.0.0 =) @mcclayton @AllPurposeName

AllPurposeName:
A release candidate branch is a great compromise. I look forward to all the breaking changes we add to it.

philipqnguyen:
lgtm

PR #128, opened 2019-01-16
Add global default field/association value configuration.

mcclayton:
Previously we only supported being able to set default values at the field/association level. This PR enables setting a global default for all fields/associations that evaluate to nil.

#### Global Config Setting
```ruby
  config.field_default = "N/A"
  config.association_default = {}
end
```

#### Field-level/Associaion-level Setting
```ruby
  identifier :uuid

  view :normal do
    field :first_name, default: "N/A"
    association :company, blueprint: CompanyBlueprint, default: {}
  end
end
```

The default value specified at the field/association level will override the value specified in the global config.

### DIFF
--- a/lib/blueprinter/extractors/association_extractor.rb
+++ b/lib/blueprinter/extractors/association_extractor.rb
@@ -14,8 +14,8 @@ module Blueprinter
 
     private
 
-    def default_value(field_options)
-      field_options.key?(:default) ? field_options.fetch(:default) : Blueprinter.configuration.association_default
+    def default_value(association_options)
+      association_options.key?(:default) ? association_options.fetch(:default) : Blueprinter.configuration.association_default
     end
 
     def association_blueprint(blueprint, value)
--- a/lib/blueprinter/extractors/association_extractor.rb
+++ b/lib/blueprinter/extractors/association_extractor.rb
@@ -6,13 +6,17 @@ module Blueprinter
 
     def extract(association_name, object, local_options, options={})
       value = @extractor.extract(association_name, object, local_options, options.except(:default))
-      return options[:default] if value.nil?
+      return default_value(options) if value.nil?
       view = options[:view] || :default
       blueprint = association_blueprint(options[:blueprint], value)
       blueprint.prepare(value, view_name: view, local_options: local_options)
     end
-    
+
     private
+
+    def default_value(field_options)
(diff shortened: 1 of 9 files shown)


lib/blueprinter/extractors/association_extractor.rb
```diff
@@ -6,14 +6,18 @@ def initialize
...
       view = options[:view] || :default
       blueprint = association_blueprint(options[:blueprint], value)
       blueprint.prepare(value, view_name: view, local_options: local_options)
     end
-    
+
     private
 
+    def default_value(association_options)
+      association_options.key?(:default) ? association_options.fetch(:default) : Blueprinter.configuration.association_default
```
AllPurposeName: Rather than this ternary, I would prefer something like
```suggestion
      assosiation_options.fetch(:default, Blueprinter.configuration.association_default)
```

What do you think?

mcclayton: I think that makes sense to me, but I avoided this initially, because I literally wanted to do a check on whether the `:default` key was present in the options in case it was explicitly set to nil, we'd want to actually render nil instead of pass through to the global Blueprinter default.
Make sense? I don't know if that's an important enough behavior to preserve or not. Thoughts? 🤔

AllPurposeName: Yeah that totally makes sense. I could see someone switching it up like that if only one or two endpoints ignore the default and use nil instead.

spec/integrations/shared/base_render_examples.rb
```diff
@@ -79,28 +79,55 @@ def extract(field_name, object, _local_options, _options={})
...
+        end
+        it('global default value is rendered for nil field') { should eq(result) }
+      end
+
+      context "Given default field value is provided" do
+        let(:result) { '{"first_name":"Unknown","id":' + obj_id + '}' }
+        let(:blueprint) do
+          Class.new(Blueprinter::Base) do
+            field :id
+            field :first_name, default: "Unknown"
```
AllPurposeName: Could we switch this example to be `nil` like the one we discussed above. Or alternatively add another test specifically for that functionality.

mcclayton: Yeah that's a great suggestion. I'll add another test case for that 👍

PR #129, opened 2019-01-18
Automate gem release process. [W.I.P.]

mcclayton:
Work in progress for automating our Gem release process via CircleCI.


philipqnguyen:
What happens when a merge to master occurred, but with no version bump?

mcclayton:
@philipqnguyen The release script will be run on every merge to master, but the script is safe in that if that version of Blueprinter is already released, it will not be re-released or overwritten. Here is a screenshot of what it will look like in CircleCI when a PR is merged into master with no version bump.

screen shot 2019-01-22 at 10 06 56 am

PR #131, opened 2019-01-22
Add field_name to if/unless procs

mcclayton:
Since the global `:if` and `:unless` procs are called at field-evaluation time, there's no way to see what field is being evaluated in the proc. So there'd be no way to accomplish #124 without passing in the field_name to do:
```ruby
  config.if = ->(field_name, obj, _options) { !obj[field_name].nil? }
  config.unless = ->(field_name, obj, _options) { obj[field_name].nil? }
end
```

# Deprecation Notice
* ⚠️ [DEPRECATION] :if/:unless procs with two arguments will now be deprecated. *These procs will now take in three arguments (field_name, obj, options) instead of just (obj, options).*
  In order to be compliant with the the next major release, all conditional :if/:unless procs must be augmented to take in three arguments instead of two. i.e. `(obj, options)` to `(field_name, obj, options)`.


hugopeixoto:
Instead of just having a breaking change in 1.0.0, would it be possible to do something like this in a 0.x.y:

```ruby
def callable_from(condition)
   callable = old_callable_from(condition)

  if callable.arity == 2
    warn "[deprecation] #{condition} now expects 3 arguments instead of 2. See <URL> for more information"
    ->(_field_name, obj, options) { callable.call(obj, options) }
  else
    callable
  end
end
```

This would allow users to upgrade and fix the deprecations before the actual breaking change, and when 1.0.0 is released, users already had the chance to fix it.

philipqnguyen:
@hugopeixoto's suggestion +1000

ritikesh:
Should do this in the last release in this 0.x series before releasing the major version.

mcclayton:
Great suggestion @hugopeixoto, I'll make those amendments / add tests to this PR, base this into master, and will create a separate PR for removing the deprecated path that will go into the `1.0.0` release candidate.

mcclayton:
I update this PR to be based into `master` instead of the `1.0.0` release candidate and incorporated @hugopeixoto's suggestion for backwards compatibility and the deprecation notice. 👍 Should be good for another round of review.

Issue #130, opened 2019-01-19
Bug: Serializing boolean `false` value returns `nil`


samsongz:
hey team! we found a bug in the serialization of boolean `false` values as follows:

To recreate:

```ruby
TestClass = Class.new(Blueprinter::Base) do
  identifier :id
  field :active
end

data = {
  id: 1,
  active: false
}

TestClass.render(data)

```
The output of `.render` is: `"{\"id\":1,\"active\":null}"`

It is coming from the `extract` method of `HashExtractor`
https://github.com/procore/blueprinter/blob/f2415a3fa0ec835ac12cf39d92974eae14ca879b/lib/blueprinter/extractors/hash_extractor.rb#L1-L7

Because the first condition evaluates to `false` (the correct value to be returned) it evaluates the second, which looks for a string key that doesn't exist, returning `nil`.

I have a branch with a spec to recreate and a suggested fix (remove the second OR condition) that I'd be happy to open a PR for to discuss or get feedback ✌️

mcclayton:
Closing this as it has been fixed in https://github.com/procore/blueprinter/pull/132

PR #132, opened 2019-01-24
[Bugfix] Serialize boolean false values

samsongz:
Proposed fix for #130 

Currently blueprinter does not serialize boolean `false` values, returning `null` instead

### Change log
- Add a test to confirm the behavior is absent in the existing code, add a boolean attribute to the specs and add a specific context block for confirming the behavior of various data types
- Remove the offending OR operation, which appears unnecessary as there does not appear to be any type checking of object keys elsewhere.

### DIFF
--- a/lib/blueprinter/extractors/hash_extractor.rb
+++ b/lib/blueprinter/extractors/hash_extractor.rb
@@ -1,7 +1,7 @@
 module Blueprinter
   class HashExtractor < Extractor
-    def extract(field_name, object, local_options, options = {})
-      object[field_name] || object[field_name.to_s]
+    def extract(field_name, object, _local_options, _options = {})
+      object[field_name]
     end
   end
 end
--- a/spec/integrations/shared/base_render_examples.rb
+++ b/spec/integrations/shared/base_render_examples.rb
@@ -10,6 +10,20 @@ shared_examples 'Base::render' do
     it('returns json with specified fields') { should eq(result) }
   end
 
+  context 'Given blueprint has ::field with all data types' do
+    let(:result) { '{"active":false,"birthday":"1994-03-04","deleted_at":null,"first_name":"Meg","id":' + obj_id + '}' }
+    let(:blueprint) do
+      Class.new(Blueprinter::Base) do
+        field :id # number
+        field :first_name # string
+        field :active # boolean
+        field :birthday # date
+        field :deleted_at # null
+      end
+    end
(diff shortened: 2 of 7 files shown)


spec/integrations/shared/base_render_examples.rb
```diff
@@ -10,6 +10,20 @@
     it('returns json with specified fields') { should eq(result) }
   end
 
+  context 'Given blueprint has ::field with all data types' do
```
samsongz: I wanted to demonstrate the failing boolean test first, but I'd be happy to extend this spec to ensure we have coverage of array and hash field data types as well

lib/blueprinter/extractors/hash_extractor.rb
```diff
@@ -1,7 +1,7 @@
 module Blueprinter
   class HashExtractor < Extractor
-    def extract(field_name, object, local_options, options = {})
-      object[field_name] || object[field_name.to_s]
+    def extract(field_name, object, _local_options, _options = {})
+      object[field_name]
```
philipqnguyen: Thanks for noticing this problem! 

However this may introduce another issue. So the original intent for `object[field_name] || object[field_name.to_s]` was to access a hash key that is a String if a Symbol key did not exist.

Can we account for String key access as well?

samsongz: I looked for other evidence (beyond this method) that string keys were explicitly supported by the library and couldn't find any in either the public documentation or the tests, please feel free to point out if I've missed something I haven't spent a ton of time w/ this gem. 

if string keys are currently supported then it appears to be an implicit and untested byproduct of the code rather than an explicit feature. I could not find any tests that have string field keys and the comments in `base.rb` https://github.com/procore/blueprinter/blob/64546aa3fe137c28d8209be0cad8e658ca92a74e/lib/blueprinter/base.rb#L24 suggest that the `method` field should be a symbol.

if that's an expected piece of functionality it feels like following this up to explicitly adding tests or documentation would be the right way to maintain it

philipqnguyen: Yes, I realized this yesterday too that i did not have any tests for String keys. I originally created the `HashExtractor` with the intent of accessing both String and Symbol keys, but I only added tests for Symbol keys. That was my oversight. 

OK, I'm fine with dropping the implicit support of the String key access in this PR.

Another PR can be made afterwards to explicitly support String keys.

PR #135, opened 2019-02-07
Add support to render with a root key and meta attributes

ritikesh:
1. Since Most JSON specs allow a root key to be configurable, we should allow that as well.
2. Added option to support passing meta information along with the root element


ritikesh:
@mcclayton I've got another one for you guys!

lib/blueprinter/base.rb
```diff
@@ -167,6 +167,11 @@ def self.association(method, options = {}, &block)
     # @option options [Symbol] :view Defaults to :default.
     #   The view name that corresponds to the group of
     #   fields to be serialized.
+    # @option options [Symbol|String] :root Defaults to nil.
```
AllPurposeName: This is not enforced, currently `[Any] :root` is more accurate.

lib/blueprinter/base.rb
```diff
@@ -225,22 +240,25 @@ def self.render_as_json(object, options= {})
...
                          view_name: view_name,
                          local_options: local_options)
         end
       else
-        object_to_hash(prepared_object,
+        data = object_to_hash(prepared_object,
                        view_name: view_name,
                        local_options: local_options)
       end
+      return data unless root
```
AllPurposeName: These three lines feel a bit sloppy and tacked on. `.prepare` is now doing potentially 7 separate things. I wouldn't feel comfortable merging something like this, but I would be more comfortable with making `.prepare` more functional and declarative... something like
```
self.prepare(args)
  raise view_not_defined unless view_defined?
  prepend_meta(prepend_root(shape_data(include_associations(args))))
end
```

I would much prefer several no-ops to complicated control flow in a bloated method.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.13.0  - 2018/12/20
```
mcclayton: Think you mean `2019/2/7` here

lib/blueprinter/base.rb
ritikesh: Wouldn't `[Any]` refer to any object whereas, we would only allow for a symbol / string as a hash key? Documentation should reflect on what is the expected input, I wouldn't want to add a `is_a` check in the method for that sake.

philipqnguyen:
@ritikesh thank you for another PR!

I wonder if we can specify root key and/or meta data in the blueprint class itself?

Maybe root key would be specified like this?

```ruby
root :users
fields :first_name, last_name, :email
```
Do you agree?

As for meta, I'm not quite sure I like the meta being passed in to `render` as is as well.

philipqnguyen:
On second thought, passing in root and meta to `render` provides better control... Scratch my previous comment =D

ritikesh:
@philipqnguyen I agree. That was the intention behind passing them to render. Better control and easier manipulation. Apart from plural differences (user for singular, users for plural), for us root element is different in different use-cases as well.

lib/blueprinter/base.rb
AllPurposeName: I'd rather deliberately decide to be open to different input or not. A type check is completely reasonable if we want to restrict to symbol/string. Is there a reason you don't want to enforce typing here?

A type coercion does happens regardless in `.render` and `.render_as_json`, but not in `.render_as_hash`. This is because the underlying `.as_json` and `.jsonify` turn `{ 1 => "hello" }` into `{ "1" => "hello"}`. I think it's confusing and dishonest to allow that to happen while `.render_as_hash`'s underlying `.prepare_for_render` leaves us with the original `{ 1 => "hello" }`.

I'm torn here because I don't think it would be wise for anyone to pass in an array, hash, or any random object, but technically now they can and it will not blow up. In those cases I would prefer to either document and fully support `any` input or restrict the input and treat it as an error and handle it gracefully by returning an instructive error message. One of my biggest pains migrating to AMS v0.10 was the lack of helpful error messages and confusion about the breaking changes surrounding `root` behavior specifically...

So let's look at some of the behavior this introduces:
```ruby
# Hash as root with render_as_hash
# Expected?
UserBlueprint.render_as_hash(
  user.where(id: 1, name: "wat"),
  root: {"My object as a root" => "yeah it's crazy"}
)

{
  {"My object as a root" => "yeah it's crazy"} => {
     "id": 1,
     "name": "wat"
  }
}
```

```ruby
# Hash as root with vanilla render
# Unexpected behavior of object becoming a string and then a key in the response.
UserBlueprint.render(
  User.where(id: 1, name: "wat"),
  root: {"My object as a root" => "yeah it's crazy"}
)

{
  "{\"My object as a root\"=>\"yeah it's crazy\"}": {
    "id": 1,
    "name": "wat"
  }
}
```

```ruby
# Integer as root with render_as_hash
# Expected because it's a hash and maintains type
UserBlueprint.render_as_hash(
  User.where(id: 1, name: "wat"),
  root: 1
)

{
  1: {
    "id": 1,
    "name": "wat"
  }
}
```

```ruby
# Integer as root with vanilla render
# Possibly unexpected coercion rather than an exception being thrown
UserBlueprint.render(
  User.where(id: 1, name: "wat"),
  root: 1
)

{
  "1": {
    "id": 1,
    "name": "wat"
  }
}
```

ritikesh: > I think it's confusing and dishonest to allow that to happen

Overall, I understand where you are coming from, but I would disagree here, because the method definitions/documentations both suggest a `json` or `json like` output - which is pretty much self-explanatory to have strings as keys in the output.

> Is there a reason you don't want to enforce typing here?

Adding extra `.is_a?(Symbol) || .is_a?(String)` for every render call does seem like an unnecessary overhead for something that should be the implementers' job. I'm however, okay with changing the type to `Any` in the comments. 

> I'm torn here because I don't think it would be wise for anyone to pass in an array, hash, or any random object

However, I still feel strongly about being indicative and not strict in this case - that you need to pass a symbol/string - This would cover most use-cases anyway.

ritikesh: On the contrary, I agree with you, we need to have proper documentation and exceptions being raised when those are not followed. Have made the changes along with some refactoring. Please let me know if you have any other concerns.

philipqnguyen: @AllPurposeName I think its ok to have it as `[Symbol|String]` because that is what we want to support. We don't want to support `[Any]` even though it is technically possible in this implementation.

Users can of course pass in Array or Hash or some other crazy type, and they would do so at their own risk. If we don't document passing in other types, we will not be responsible for breaking it in future updates.

ritikesh: @AllPurposeName, @philipqnguyen, I've retained the `Symbol | String` documentation and have added a check to raise exception if anything else is passed for root along with a test case for the same. 

I agree with @AllPurposeName that it'd be counter intuitive not raising an exception when calling out that limited types are supported.

Along with that, I've done some other refactoring as well. The base class was becoming heavier with lot of helper methods and active record helper has not grown as it was anticipated. Hence, I have combined them into one and have moved all private helper methods there. You can refer to my most recent commit. 

Let me know your thoughts.

ritikesh:
Hi guys, any updates on this?

lib/blueprinter/helpers/base_helpers.rb
```diff
@@ -0,0 +1,101 @@
+module Blueprinter
```
AllPurposeName: 💯 	this is a great step in organization!

lib/blueprinter/helpers/base_helpers.rb
```diff
@@ -0,0 +1,101 @@
...
+      private
+      def active_record_relation?(object)
+        !!(defined?(ActiveRecord::Relation) &&
+          object.is_a?(ActiveRecord::Relation))
+      end
+
+      def prepare_for_render(object, options)
+        view_name = options.delete(:view) || :default
+        root = options.delete(:root)
+        symbol_or_string_or_nil?(root, :root)
```
AllPurposeName: Not blocking: Not a fan of this method name, but I like what it does!

spec/integrations/shared/base_render_examples.rb
```diff
@@ -287,6 +287,37 @@ def self.unless_method(_object, _options)
...
+    let(:blueprint) { blueprint_with_block }
+    it('returns json with a root') do
+      expect(blueprint.render(obj, root: :root, meta: 'meta_value')).to eq(result)
+    end
+  end
+
+  context 'Given blueprint has ::meta without root' do
+    let(:result) { '{"id":' + obj_id + ',"position_and_company":"Manager at Procore"}' }
+    let(:blueprint) { blueprint_with_block }
+    it('returns json with a root') do
```
mcclayton: Is this description accurate? Shouldn't this be `returns json without root and meta` since a root is not provided?
Should this just drop the meta or should it throw a Blueprinter error since a root was not specified?

ritikesh: The implementers would clearly know from responses generated. Moreover, we have explicitly called this out in both the readme and the documentation/comments in the methods. Any deviations in future from this behaviour would be caught by this Spec.

AllPurposeName: I don't think we should rely on the implementers noticing something isn't showing up. I would not be satisfied with the answer "read the documentation better" when something is failing silently on me.

PR #139, opened 2019-03-02
remove include_associations

ritikesh:
Based on our discussions on gitter, removing include_associations


mcclayton:
Not sure why the CircleCI checks aren't showing up in Github, but confirmed they are passing from the CircleCI console so going ahead and merging/releasing.

PR #141, opened 2019-03-21
exclude multiple fields inline

pabhinaya:
@mcclayton , For derived views with multiple included views, ability to exclude multiple fields inline.

### DIFF
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -309,6 +309,29 @@ module Blueprinter
     def self.exclude(field_name)
       current_view.exclude_field(field_name)
     end
+    
+    # When mixing multiple views under a single view, some fields may required to be excluded from
+    # current view
+    # 
+    # @param [Array<Symbol>] the fields to exclude from the current view.
+    #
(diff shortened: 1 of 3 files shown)


mcclayton:
Yes, if you wouldn't mind adding a test to `view_spec.rb` similar to: https://github.com/procore/blueprinter/blob/master/spec/units/view_spec.rb#L15-L23
and then I can handle drafting up the release/version bump/CHANGELOG entry.

pabhinaya:
@philipqnguyen, @mcclayton  I have added the specs for exclude_fields. Please proceed.

ritikesh:
I believe exclude should be able to take 1 or more arguments. Both the below examples should be valid usages of exclude:

```ruby
exclude :field1
```

```ruby
exclude :field1, :field2
```

This should be possible with Ruby's splat operator.

philipqnguyen:
@ritikesh normally I would agree with your suggestion too, but I think having plural `excludes` would fit with current patterns: ie.,`field` for single and `fields` for multiple fields.

PR #142, opened 2019-03-26
Allow either string or procs to be used in formatting DateTime objects

mcclayton:
This aims to address https://github.com/procore/blueprinter/issues/120

Previously, we had a field level option `datetime_format` which takes in a string representing the strptime format to format the Date/DateTime object. This PR enables `datetime_format` to be either a String *or* a Proc. This enables more flexibility, especially for those who want to have something like integer UNIX timestamps.

### Usage (String Option):
```ruby
  identifier :name
end
```
Output:
```json
{
  "name": "John Doe",
  "birthday": "03/04/1994"
}
```

### Usage (Proc Option):
```ruby
  identifier :name
  field :birthday, datetime_format: ->(datetime) { datetime.nil? ? datetime : datetime.strftime("%s").to_i }
end
```

Output:
```json
{
  "name": "John Doe",
  "birthday": 762739200
}
```


mcclayton:
Actually, going to close this for now — not quite ready.
This is something that can already be achieved by passing a block directly to the field, however, I am viewing this as a stepping stone towards having a global datetime_format, where you’d simply be able to have `datetime_format: true` to use the global formatter, datetime_format: ->datetime { datetime.year.to_s } for when you want to override the global with a proc, and same for the string version. I'll reopen this PR once I have the global formatting instrumented, and after confirming this approach with @philipqnguyen.

PR #143, opened 2019-03-27
Global datetime formatter

ritikesh:


spec/integrations/shared/base_render_examples.rb
```diff
@@ -64,30 +64,23 @@ def extract(field_name, object, _local_options, _options={})
...
         identifier :id
-        field :birthday,   datetime_format: "%m/%d/%Y"
+        field :birthday
         field :deleted_at, datetime_format: '%FT%T%:z'
       end
     end
     it('returns json with a formatted field') { should eq(result) }
   end
 
-  context 'Given blueprint has a :datetime_format argument on an invalid ::field' do
```
mcclayton: I'd keep but modify this test. Something more along the lines of:
```
context 'Given blueprint has a :datetime_format argument on an non-date ::field' do
   ...
   it('does not apply the date format')...
end
```

lib/blueprinter/formatters/date_time_formatter.rb
```diff
@@ -0,0 +1,18 @@
+module Blueprinter
+  class DateTimeFormatter
+    def extract(value, options)
+      if value.respond_to?(:strftime)
+        value = format_datetime(value, options)
+      end
+      value
```
philipqnguyen: I think we need to alter the logic here a bit in order to prevent a breaking change.
1. If `options[:datetime_format]` is passed and `!value.respond_to?(:strftime)`, we should raise the original `BlueprinterError`.
2. Otherwise, follow through with the rest of this logic.

lib/blueprinter/extractors/auto_extractor.rb
```diff
@@ -4,11 +4,13 @@ def initialize
...
       @public_send_extractor = PublicSendExtractor.new
       @block_extractor = BlockExtractor.new
+      @datetime_formatter = DateTimeFormatter.new
     end
 
     def extract(field_name, object, local_options, options = {})
       extraction = extractor(object, options).extract(field_name, object, local_options, options)
-      value = options.key?(:datetime_format) ? format_datetime(extraction, options[:datetime_format]) : extraction
+      
+      value = @datetime_formatter.extract(extraction, options)
```
philipqnguyen: In order to follow the current naming paradigm, how do you feel if we renamed `DateTimeFormatter#extract` to `DateTimeFormatter#format`? Since `#extract` makes more sense with the `Extractor` classes, but not so for this Formatter class.

PR #145, opened 2019-04-01
Allow either string or procs to be used in formatting DateTime objects

mcclayton:
This aims to address https://github.com/procore/blueprinter/issues/120

Previously, we had a global and field level option `datetime_format` which takes in a string representing the strptime format to format the Date/DateTime object. This PR enables `datetime_format` to be either a String *or* a Proc. This enables more flexibility/control in how dates can be formatted. ( i.e. for those who want to have something like integer UNIX timestamps)

This is something that can _technically already be achieved (only at the field-level) _ by passing a block directly to the field, however, now that datetime_format can be set globally, this allows us to set a global datetime_format Proc which will get invoked on any field that responds to `strptime`, and matching field-level syntax in a more idiomatic way. 

## Global Config Setting
If a global datetime_format is set (either as a string format or a Proc), this option will be
invoked and used to format all fields that respond to `strptime`.
```ruby
  config.datetime_format = ->(datetime) { datetime.nil? ? datetime : datetime.strftime("%s").to_i }
end
```

### Usage (String Option):
```ruby
  identifier :name
end
```
Output:
```json
{
  "name": "John Doe",
  "birthday": "03/04/1994"
}
```

### Usage (Proc Option):
```ruby
  identifier :name
  field :birthday, datetime_format: ->(datetime) { datetime.nil? ? datetime : datetime.strftime("%s").to_i }
end
```

Output:
```json
{
  "name": "John Doe",
  "birthday": 762739200
}
```


README.md
```diff
@@ -524,6 +538,22 @@ Output:
...
+```
+
+Output:
+```json
+{
+  "name": "John Doe",
+  "birthday": 762739200
+}
+```
+
```
AllPurposeName: Might be nice to add a clarifying line such as what is on [Line 506](https://github.com/procore/blueprinter/pull/145/files#diff-04c6e90faac2675aa89e2176d2eec7d8R506)

lib/blueprinter/formatters/date_time_formatter.rb
```diff
@@ -16,7 +16,16 @@ def format(value, options)
 
     def format_datetime(value, field_format)
       format = field_format || Blueprinter.configuration.datetime_format
-      format.nil? ? value : value.strftime(format)
+
+      if format.nil?
```
AllPurposeName: I would rather see this as a case statement since each branch is based on type.

lib/blueprinter/formatters/date_time_formatter.rb
```diff
@@ -16,7 +16,16 @@ def format(value, options)
...
+        value
+      elsif format.is_a?(Proc)
+        format.call(value)
+      elsif format.is_a?(String)
+        value.strftime(format)
+      else
+        raise BlueprinterError, 'Cannot format DateTime object with invalid formatter'
```
AllPurposeName: We ought to raise a more specific error since this is exclusively in the date time formatter domain.

spec/integrations/shared/base_render_examples.rb
```diff
@@ -91,6 +91,50 @@ def extract(field_name, object, _local_options, _options={})
...
+  end
+
+  context 'Given blueprint has a Proc :datetime_format argument on an invalid ::field' do
+    let(:blueprint) do
+      Class.new(Blueprinter::Base) do
+        identifier :id
+        field :first_name, datetime_format: -> datetime { datetime.strftime("%s") }
+      end
+    end
+    it('raises a BlueprinterError') { expect{subject}.to raise_error(Blueprinter::BlueprinterError) }
```
AllPurposeName: Hmm the error here is based on us using a datetime formatter on an invalid field, correct? The error message we get reads `'Cannot format DateTime object with invalid formatter'` The formatter in this test case is perfectly fine, no? I think we need an invalid field error or something.

AllPurposeName: Nevermind I was confused. The message *DOES NOT* read `'Cannot format DateTime object with invalid formatter'`, rather it's `'Cannot format invalid DateTime object'` which is perfectly valid.

lib/blueprinter/formatters/date_time_formatter.rb
AllPurposeName: Also maybe a message such as `"Cannot format DateTime object with invalid formatter: #{format.class}"`

PR #149, opened 2019-05-02
Exclude a view functionality

Abhinaya-P:
Though excludes can be used to exclude several fields. If all the excluded fields are from a single view, then exclude_view is a more refined and usable option

@mcclayton , @philipqnguyen

### DIFF
--- a/lib/blueprinter/view_collection.rb
+++ b/lib/blueprinter/view_collection.rb
@@ -51,7 +51,7 @@ module Blueprinter
 
       views[view_name].excluded_view_names.each do |excluded_view_name|
         next if view_name == excluded_view_name
-        fields.except(views[excluded_view_name].fields.keys)
+        fields.except!(*views[excluded_view_name].fields.keys)
       end
 
       fields
--- a/lib/blueprinter/view_collection.rb
+++ b/lib/blueprinter/view_collection.rb
@@ -49,6 +49,11 @@ module Blueprinter
         fields.delete(name)
       end
 
+      views[view_name].excluded_view_names.each do |excluded_view_name|
+        next if view_name == excluded_view_name
+        fields.except(views[excluded_view_name].fields.keys)
+      end
+
       fields
     end
 
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -293,6 +293,44 @@ module Blueprinter
       current_view.include_view(view_name)
     end
 
+    # Specify a view that needs to be excluded from the current view.
+    #
+    # @param view_name [Symbol] the view to exclude from the current view.
+    # PURPOSE: Though excludes can be used to exclude several fields. If all the excluded fields are from a single view, then exclude_view
+    # is a more refined and usable option
+    #
+    # @example Excluding a view from a dervied view
+    #   class UserBlueprint < Blueprinter::Base
+    #     # other code...
+    #     view :normal do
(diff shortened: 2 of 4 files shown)


mcclayton:
Thank you so much for contributing @Abhinaya-P, I'll try to get some time soon to take a look at this, but I'll be on vacation soon. May have to check it out when I get back next week.

lib/blueprinter/view_collection.rb
```diff
@@ -49,6 +49,11 @@ def sortable_fields(view_name)
         fields.delete(name)
       end
 
+      views[view_name].excluded_view_names.each do |excluded_view_name|
```
mcclayton: I don't believe this is properly excluding all fields from a specified `exclude_view`.

May be worth adding this test into `base_render_examples`:

```ruby
  context 'Given blueprint has excluded ::views' do
    let(:exclude_minimal) do
      ['{"id":' + obj_id + '', '"description":"A person"',
      '"employer":"Procore"', '"position":"Manager"}'].join(',')
    end
    let(:exclude_normal) do
      ['{"id":' + obj_id + '', '"description":"A person"}'].join(',')
    end
    let(:blueprint) do
        identifier :id
        view :minimal do
        end
        view :normal do
          include_view :minimal
          fields :position
          field :company, name: :employer
        end
        view :extended do
          include_view :normal
          field :description
        end
        view :exclude_minimal do
          include_view :extended
          exclude_view :minimal
        end

        view :exclude_normal do
          include_view :extended
          exclude_view :normal
        end
      end
    end
    it('returns json with view fields excluded') do
      expect(blueprint.render(obj, view: :exclude_minimal)).to eq(exclude_minimal)
      expect(blueprint.render(obj, view: :exclude_normal)).to eq(exclude_normal)
    end
  end
```
You'll notice this test will fail for this code as `exclude_view :normal` won't actually exclude all the fields in `normal`.

Abhinaya-P: @mcclayton , when default views come into picture, I believe it expected to be always included in the final render regardless of any exclude_views (containing the default view). can you share your thoughts on this.

philipqnguyen: IMO default view should always be included as you suggested @Abhinaya-P unless `exclude_view :default` is specifically specified.

mcclayton: Ahh yes I agree that default view should always be included (also unless `exclude_view :default`) is specifically provided 👍

Abhinaya-P: @mcclayton - I have included the functionality to handle default views and added few more tests. A thorough review and testing would be helpful here. Thanks.

mcclayton:
Hi @Abhinaya-P,
Thank you for both your contribution to Blueprinter and your patience with us as we look into this.
To provide some insight, there are a couple things holding us up on this PR currently.
1. This PR has exposed some potential undefined behavior and edge cases in our current `exclude` field functionality
which we are looking into and trying to address before we can determine exactly how `exclude_view` should behave in the similar edge cases.
i.e. we have not clearly defined the order of operations of excluding/including fields -- and so when we have multiple views including/excluding fields, we are still determining what the order of operations/behavior should be.
2. We are still considering if `exclude_view` is functionality we want to introduce into the Blueprinter gem.

pabhinaya:
Sure. @mcclayton . Let me know when things are clear from your end. Until then this will be on hold.

PR #150, opened 2019-05-15
add mongoid support.

ayamomiji:
or maybe just use `object.respond_to?(:to_a)`?

### DIFF
--- a/lib/blueprinter/helpers/base_helpers.rb
+++ b/lib/blueprinter/helpers/base_helpers.rb
@@ -11,6 +11,11 @@ module Blueprinter
           object.is_a?(ActiveRecord::Relation))
       end
 
+      def mongoid_criteria_relation?(object)
+        !!(defined?(Mongoid::Criteria) &&
+          object.is_a?(Mongoid::Criteria))
+      end
+
       def prepare_for_render(object, options)
         view_name = options.delete(:view) || :default
         root = options.delete(:root)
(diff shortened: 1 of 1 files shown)


ayamomiji:
https://www.rubydoc.info/github/mongoid/mongoid/Mongoid%2FDocument:to_a
since all mongoid documents are respond to `to_a`, the solution seems not quite appropriate :(

philipqnguyen:
thanks for the PR @ayamomiji 

`to_a` won't work out of the box because Hash objects also responds to `to_a`.

On top of that, we can't claim to support Mongoid if we don't add tests and appropriate README.

However, before you start doing anymore work on this PR (because I don't want you to waste your time), I have some concerns that should be addressed first. Namely, should Blueprinter support Mongoid? If we do, we need to maintain tests, docs, and ensure future features to be compatible for Mongoid as well. I'm actually leaning towards not supporting mongoid, but can be persuaded.

cc. @mcclayton @AllPurposeName @dlcarter @adamwells @jl4ever @nolanprocore @njbbaer @wai-procore

mcclayton:
I agree with @philipqnguyen in that, if we were to support Mongoid directly from within Blueprinter, we'd want to fully support it via documentation, testing, adding it to our compatibility checks/roadmap, etc.

mcclayton:
Thank you so much again for your contribution @ayamomiji,

mcclayton:
Thank you so much again for your contribution @ayamomiji,
Apologies, but as mentioned above, this is not something we plan to directly support as implemented here, however, I created an issue here https://github.com/procore/blueprinter/issues/153 to track/ideate/create a more extensible approach to providing configurable ORM/ODM support. In the meantime, please feel free to fork Blueprinter and use your implementation if you need this support immediately. Going to go ahead and close this PR for now.

PR #154, opened 2019-05-21
fix: Early return identifier fields if view matches.

AllPurposeName:
Before this when we passed in `view: :identifier`, we were incorrectly adding `sorted fields` to the identifier fields. Sorted fields would include all fields and associations not scoped to a specific view.

The expected behavior is that if `view == :identifier`, then only the identifier fields will be returned.

Thus,
```ruby
# Setup
class MyBlueprint < Blueprinter::Base
  identifier :id
  field :first_name

  view :normal do
    field  :last_name
  end
end
```

```ruby
# Old behavior
MyBlueprint.render(obj, view: identifier) #=> { "id": 1, "first_name": "Rock" }
```
vs
```ruby
# New behavior
MyBlueprint.render(obj, view: identifier) #=> { "id": 1 }
```


philipqnguyen:
Would this be a bugfix rather than a breaking change?

It seems to me that if I specify `:identifier` but I get in return fields outside of `:identifier` then it is a bug, right?

mcclayton:
@philipqnguyen It's certainly a bug fix, but we were thinking this may be a breaking bug fix as we cannot guarantee that any users are not referencing identifier view fields exposed as part of the bug.

philipqnguyen:
Then would every bugfix be breaking? Since we cannot guarantee that people weren't using a bugged feature?

dlcarter:
I agree with @philipqnguyen that this is a bug fix, and not a breaking change. I can't imagine anyone rightfully running into this bug and _relying_ on it.

mcclayton:
Discussed more offline, we determined that this is a bugfix, and not a `1.0.0` release candidate. In `1.0.0`, we will determine whether or not we even want to keep the Blueprinter defined `identifier` view around.

AllPurposeName:
Changed the branch, I will resolve conflicts.

PR #156, opened 2019-05-24
Dont run benchmarks for ruby 2.2

philipqnguyen:
Ruby 2.2 performance is not reliable, and it regularly fails the
blueprinter benchmark. We still support ruby 2.2 officially, but
can't be confidence in its performance.

We still are running benchmarks on 2.3, 2.4, and 2.5

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 
 steps: &steps_spec
   - checkout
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
```
philipqnguyen: I'm using the `when` conditional, it requires version 2.1

.circleci/config.yml
```diff
@@ -7,23 +7,42 @@ steps: &steps_spec
...
+        - run:
+            command: bundle exec rake benchmarks
 
 jobs:
-  ruby-2.2:
+  ruby-2_2:
+    parameters:
+      benchmark:
+        type: boolean
+        default: false
```
philipqnguyen: passing in benchmark parameters here to false, while the other ruby versions will be true.

.circleci/config.yml
```diff
@@ -7,23 +7,42 @@ steps: &steps_spec
       command: bundle check --path=vendor/bundle || bundle install --path=vendor/bundle
   - run:
       command: bundle exec rake
-  - run:
-      command: bundle exec rake benchmarks
+  - when:
+      condition: << parameters.benchmark >>
+      steps:
+        - run:
+            command: bundle exec rake benchmarks
```
philipqnguyen: runs benchmark if paramenters.benchmark is true

.circleci/config.yml
```diff
@@ -7,23 +7,42 @@ steps: &steps_spec
...
+  ruby-2_4:
+    parameters:
+      benchmark:
+        type: boolean
+        default: true
     steps: *steps_spec
     docker:
       - image: ruby:2.4
-  ruby-2.5:
+  ruby-2_5:
```
philipqnguyen: circleci version 2.1 does not allow `.`, so I replaced it with `_`

adamwells: This won't be an issue on the version key at the top of this file?

adamwells: Ah, I see, only keys and not values?

philipqnguyen: @adamwells yes!

.circleci/config.yml
```diff
@@ -7,23 +7,42 @@ steps: &steps_spec
       command: bundle check --path=vendor/bundle || bundle install --path=vendor/bundle
   - run:
       command: bundle exec rake
-  - run:
-      command: bundle exec rake benchmarks
+  - when:
+      condition: << parameters.benchmark >>
```
mcclayton: Nice, I like this approach to still run tests against it, but not benchmarks 👍

PR #163, opened 2019-06-28
Dynamic views functionality

fpaint:
I changed include_view so it can receive a block to evaluate the view name from object and options. It also works for associations. If association view option is a proc then the view name is evaluated.

### DIFF
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -248,6 +248,7 @@ module Blueprinter
     #
     # @api private
     def self.prepare(object, view_name:, local_options:, root: nil, meta: nil)
+      view_name = view_name.is_a?(Proc) ? view_name.call(object, local_options) : view_name  
       unless view_collection.has_view? view_name
         raise BlueprinterError, "View '#{view_name}' is not defined"
       end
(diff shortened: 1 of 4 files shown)


lib/blueprinter/base.rb
```diff
@@ -292,8 +293,8 @@ def self.fields(*field_names)
     #   end
     #
     # @return [Array<Symbol>] an array of view names.
-    def self.include_view(view_name)
-      current_view.include_view(view_name)
+    def self.include_view(view_name = nil)
+      current_view.include_view(view_name || Proc.new)
```
philipqnguyen: Above this method, would you add comments to note that it optionally accepts a proc? As well as an example?
For example, `#field` has the following comments: https://github.com/procore/blueprinter/blob/4ba12dd14edf340f9a4a95f1668785adc8c21ca6/lib/blueprinter/base.rb#L87-L102

These comments are parsed by yard and uploaded to rubydocs which produces documentation like this: https://www.rubydoc.info/github/procore/blueprinter/Blueprinter/Base#field-class_method

Additionally, would you update the readme in the "View" section to give an example of passing a block to `include_view`

dlcarter:
There are a number of other related PRs and concerns we are taking into consideration. We are looking into how to best consolidate these approaches into one implementation. This effort can be tracked with https://github.com/procore/blueprinter/projects/2#card-57792153

PR #164, opened 2019-07-04
Add dynamic field support

amalarayfreshworks:
We have a use case in our product where the key names of the result hash are defined at run time, so in order to avoid introducing an unnecessary extra level/key, I have made some changes.

I have introduced one custom field option called "merge". If this option is present we can add the value returned by the "method/name" directly to the obj hash through a merge. 

Eg :
let(:blueprint) do
        identifier :id
        field :dynamic_fields, merge: true
      end
end
def dynamic_fields
  {"field1" => "val1"}
end
 blueprint.render(obj)  would return '{"id":'  + obj_id + ',  "field1":  "val1"}'
 instead of 
 '{"id":'  + obj_id + ',  "dynamic_fields": {"field1":  "val1"}}'


spec/factories/model_factories.rb
```diff
@@ -3,7 +3,7 @@
 FactoryBot.define do
   factory :user do
     first_name { 'Meg' }
-    last_name  { 'Jones' }
+    last_name  { 'Ryan' }
```
AllPurposeName: 💯

amalarayfreshworks:
@philipqnguyen will make those changes and push it

philipqnguyen:
@amalarayfreshworks I'm not sure if I fully understand.

In this case:
```ruby
field :something, name: -> { |obj, _opt| obj.dynamic_field_name }
```
The `field :something` would be responsible for accessing the value, where as `name: -> { |obj, _opt| obj.dynamic_field_name }` would evaluate to a new name. I don't see the `name` proc ever evaluating for the name and value.

Or am I missing something or not understanding you?

ritikesh:
@philipqnguyen can you pls check Gitter? She has elaborated the use-case there.

amalarayfreshworks:
I have elaborated the usecase in gitter https://gitter.im/blueprinter-gem/community?at=5d24b57317cc7b05ca9c8249. Please do let me know how can we implement the use case by being in line with current blueprinter's APIs.

mcclayton:
The team has been deliberating on how we’d like to address something like this.
We’ve come up with a proposal that we think is generic enough to handle your use case, as well as some others that we’ve been thinking through (such as https://github.com/procore/blueprinter/issues/158). Would the following suffice your use case?

We are thinking of adding a new Transformer class
i.e.

```
module Blueprinter
  class Transformer
    def transform(blueprinter_obj, object, options={})
      fail NotImplementedError, "A Transformer must implement #transform"
    end

    def self.transform(blueprinter_obj, object, options={})
      self.new.transform(blueprinter_obj, object, options)
    end
  end
end
```


Users can then implement transformer classes and declare them in a Blueprint as follows:
```

  transform DynamicFieldTransformer
end

class DynamicFieldTransformer < Blueprinter::Transformer
  def transform(blueprinter_obj, object, options)
    blueprinter_obj.merge!(object.dynamic_fields)
  end
end
```

Whatever is returned by the set `transform` will be the resulting hash passed to serialization (to_json).

amalarayfreshworks:
Can anyone please re trigger the build for ruby 2_5 and ruby 2_4 as the build at installing ruby dependencies. 
image

lib/blueprinter/transformer.rb
```diff
@@ -0,0 +1,12 @@
...
+  class Transformer
+    def transform(result_hash,primary_obj, options={})
+      fail NotImplementedError, "A Transformer must implement #transform"
+    end
+
+    def self.transform(result_hash,primary_obj, options={})
+      self.new.transform(result_hash,primary_obj, options)
+    end
+  end
+end
```
philipqnguyen: would you add a newline at end of file? github complains about that

lib/blueprinter/view.rb
```diff
@@ -38,6 +43,10 @@ def exclude_fields(field_names)
       end
     end
 
+    def *(custom_transformer)
+      transformers << custom_transformer
+    end
```
philipqnguyen: I've never seen `*` used for anything but multiplication. Is this common style?

Otherwise, can we come up with a method name that makes clear that it is appending transformer here?
I'm worried we will run out of symbols to define new methods 🤣

amalarayfreshworks: I am usually not into operator overloading, but then I was inspired by the method written for adding fields into the view
def <<(field)
      fields[field.name] = field
end
I was under the impression that its a style blueprinter follows for adding setters. I agree to all the reviewers @philipqnguyen  and @mcclayton  when you feel that we can run out of operators to overload. For now I will change the method name to transform :)

lib/blueprinter/view.rb
```diff
@@ -22,6 +23,10 @@ def inherit(view)
       view.excluded_field_names.each do |field_name|
         exclude_field(field_name)
       end
+      
+      view.transformers.each do |transformer|
+        self * transformer
```
philipqnguyen: you'll want to change this `*` to the new method name.

lib/blueprinter/view.rb
philipqnguyen: can you rename to `add_transformer`?  Otherwise it seems like `view.transform` is already invoking the transformer, which is not entirely accurate since it's just storing it to be called later.

spec/integrations/shared/base_render_examples.rb
```diff
@@ -425,4 +425,22 @@ def self.unless_method(_field_name, _object, _options)
...
+          result_hash.merge!(dynamic_fields)
+        end
+      end
+      Class.new(Blueprinter::Base) do
+        identifier :id
+        transform DynamicFieldsTransformer
+      end
+    end
+    it('returns json with values derived from options') { should eq(result) }
+  end
```
philipqnguyen: Would you also add a test to also ensure that the transformers are also inherited? There's some inheritance specs here that you can add to: https://github.com/procore/blueprinter/blame/4ba12dd14edf340f9a4a95f1668785adc8c21ca6/spec/integrations/base_spec.rb#L326-L348

amalarayfreshworks: Test case has been added.

lib/blueprinter/base.rb
```diff
@@ -274,6 +275,43 @@ def self.fields(*field_names)
...
+    #      end
+    #   end
+    #
+    #   class UserBlueprint < Blueprinter::Base
+    #     fields :first_name, :last_name
+    #     transform DynamicTransformer  
+    #     # other code
+    #   end
+    #
+    #   class DynamicFieldTransformer < Blueprinter::Transformer
```
tjwallace: Should this be `DynamicTransformer`? (or `transform DynamicFieldTransformer` above)?

philipqnguyen: Yup, example mismatch. would appreciate if you can make a PR.

tjwallace: Done, see #174

PR #165, opened 2019-07-10
Remove API private from Extractor

cagmz:
Custom `Extractor` classes are already being used in production and the Private API warning in the docs is confusing for developers.

https://www.rubydoc.info/gems/blueprinter/Blueprinter/Extractor

### DIFF
--- a/lib/blueprinter/extractor.rb
+++ b/lib/blueprinter/extractor.rb
@@ -1,5 +1,4 @@
 module Blueprinter
-  # @api private
   class Extractor
     def extract(field_name, object, local_options, options={})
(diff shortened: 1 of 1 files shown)


philipqnguyen:
@cagmz thanks for doing this, it makes sense to make the base `Extractor` public so that others can inherit from it and create their own custom extractors. However, would you mark `AssociationExtractor`, `AutoExtractor`, `BlockExtractor`, `HashExtractor`, and `PublicSendExtractor` as `@api private`? I don't think we can promise that those won't change.

cagmz:
Closing in favor of https://github.com/procore/blueprinter/pull/182 .

PR #168, opened 2019-07-15
WIP: Add AMS extractor to associate AMS resources

HParker:
Now you should be able to integrate your AMS models with your blueprints as long as the blueprint is the top level object. This should aid in migrating to blueprint without having to migrate all associated objects.

It should be very easy to add support for other `foreign_association` types in the future if people have trouble migrating from another framework.

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,6 +354,23 @@ Output:
 }
 ```
 
+Associations can also use AMS to render the association:
+
+```ruby
+class UserSerializer < ActiveModelSerializers::Model
+  attributes :id
+end
+class UserBlueprint < Blueprinter::Base
+  identifier :uuid
+  field :email, name: :login
+
+  view :normal do
+    fields :first_name, :last_name
+    foreign_association :user, serializer_type: :ams
+  end
+end
+```
+
 ---
 </details>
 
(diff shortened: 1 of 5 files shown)


HParker:
Forgot to add: Closes: https://github.com/procore/blueprinter/issues/167

HParker:
One thing I am thinking about adding here is a way to pass options to the AMS model, but I want to try it out a bit more first and see what options (if any) are actually useful.

HParker:
Changed title to WIP. I was messing around with this more and realized that there are some big issues with this approach. `instance_options` and `SerializerContext` are a bit difficult to match behavior for in Blueprinter.

I am starting to doubt the validity of this approach. It seems possible that the concepts used in AMS are too dissimilar to make an easy integration between the two.

philipqnguyen:
Just a suggestion, what do you think if we override this in our own project first before attempting a merge into blueprinter? That would give us time to iron out the edge cases?

HParker:
Yeah, my thought about this feature now is that it is a bad idea. In reality serializer usage is too individual for us to automate away how you use your serializer, people migrating would have a better time just using the block syntax on field to define how their association is serialized. Closing for now. Thanks!

PR #175, opened 2019-08-13
Moving away from hashed parameters to regular ones

ritikesh:
This is to reduce the number of object(hashes) allocations. The intended functionality can be achieved using named n-ary arguments as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.19.1  - 2019/08/13
+* 💅 [ENHANCEMENT] Moved away from hash based named parameters to reduce object allocations. [#175](https://github.com/procore/blueprinter/pull/175). Thanks to [@ritikesh](https://github.com/ritikesh).
+
 ## 0.19.0  - 2019/07/24
 * 🚀 [FEATURE] Added ability to specify transformers for Blueprinter views to further process the resulting hash before serialization. [#164](https://github.com/procore/blueprinter/pull/164). Thanks to [@amalarayfreshworks](https://github.com/amalarayfreshworks).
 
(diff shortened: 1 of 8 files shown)


ritikesh:
Simple benchmark that I tried:

```ruby
def dynamic(a: , b: , c: )
	# puts a,b,c
end

def dynamicc(a , b , c)
	# puts a,b,c
end

TESTS = 100_000
Benchmark.bmbm do |results|
	results.report('hashed') {
		TESTS.times {
			dynamic(a: 1, b: 2, c: 3)
		}
	}
	results.report('regular') {
		TESTS.times {
			dynamicc(1, 2, 3)
		}
	}
end
```

```shell
Rehearsal -------------------------------------------
hashed    0.010000   0.000000   0.010000 (  0.013016)
regular   0.010000   0.000000   0.010000 (  0.005975)
---------------------------------- total: 0.020000sec

              user     system      total        real
hashed    0.010000   0.000000   0.010000 (  0.010971)
regular   0.000000   0.000000   0.000000 (  0.006538)
```

philipqnguyen:
Not a fan of refactoring something unless it is absolutely warranted. Is there any major difference if we benchmarked using actual blueprinter?

ritikesh:
Ticket is our most common object and it is serialised at least a few million times a day. I've benchmarked only for a 1000 runs.

```ruby
TESTS = 1000
Benchmark.bmbm do |results|
	results.report() {
		TESTS.times {
			TicketBlueprint.render(tkt, view: :skeleton)
		}
	}
end
```

```shell
Current gem version(0.19.0):
Rehearsal ------------------------------------
   6.050000   0.210000   6.260000 (  6.336973)
--------------------------- total: 6.260000sec

       user     system      total        real
   5.090000   0.070000   5.160000 (  5.182255)

Forked gem version(0.19.1):
Rehearsal ------------------------------------
   5.010000   0.050000   5.060000 (  5.078678)
--------------------------- total: 5.060000sec

       user     system      total        real
   4.370000   0.060000   4.430000 (  4.442997)
```

ritikesh:
@philipqnguyen, I agree, however with good unit/integration test coverages, I see no reasons to say no to any refactor. All code(good/bad) should be periodically tuned. 😄

philipqnguyen:
@ritikesh my hesitation with this PR is that keyword arguments are very common in Ruby, and it provides an easy to read interface, especially an interface where argument ordering doesn't matter. As such I don't want to replace it just yet. I'm sure there are other parts of this gem that can be worked on that might provide similar or better performance gains.

philipqnguyen:
I'm closing this for now due to inactivity.

PR #176, opened 2019-08-19
add Rails generators

wlkrw:
Hello and thanks for making Blueprinter.

I made this for https://github.com/procore/blueprinter/issues/121

Would YARD doc additions be needed or is in this case the  ```rails g blueprinter:blueprint --help``` output sufficient?


wlkrw:
ok, I'll fix for old rubies

jesster2k10:
Hey, what's the story with this?

wlkrw:
@jesster2k10 I've been offline for a couple weeks, and I don't know that there is much of a story to tell but if you are still interested you can always bundle with my fork and play with the generator this would add. It can introspect on the ActiveRecord model you pass it or just take command flags to write a blueprinter for the model.

spec/generators/blueprint_generator_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+    describe "given --detect_associations --default_association" do
+      include_examples "generated_file"
+      before do
+        run_generator %W(#{model} --detect_associations --default_association)
+      end
+
+      it "blueprint file has default association" do
+        is_expected.to contain(/default: {}/)
+      end
+    end
```
philipqnguyen: While indeed the `{}` was used as an example in the README, it is not always the case. Devs can specify any kind of defaults for example:
```
association :friends, blueprint: FriendBlueprint, default: []
association :pictures, blueprint: PictureBlueprint, default: {}
association :posts, blueprint: PostBlueprint, default: nil
association :mice, blueprint: MouseBlueprint, default: ["Minnie Mouse", "Mickey Mouse"]
```

To be honest, I think it's OK to not support association defaults via the generator.

spec/generators/blueprint_generator_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+    describe "given --detect_associations --dynamic_association" do
+      include_examples "generated_file"
+      before do
+        run_generator %W(#{model} --detect_associations --dynamic_association)
+      end
+
+      it "blueprint file has detected association" do
+        is_expected.to contain(/blueprint: ->\(user\) {user.blueprint}/)
+      end
+    end
```
philipqnguyen: Same here, I don't think we need to support dynamic associations via the generator. While the example used in the README is `obj.blueprint` that is not always the case. The devs can do anything they want in that block, as long as it evaluates to a `Blueprinter::Base`.

lib/generators/blueprinter/blueprint_generator.rb
```diff
@@ -65,7 +65,7 @@ def identifier_symbol
 
       def fields
         fs = if options["detect_fields"]
-               [].concat(options["fields"], introspected_fields)
+               Array.new(options["fields"]).concat(introspected_fields)
```
philipqnguyen: What happens if `--fields` and `--detect_fields` overlap with the same field names? There would be duplicates right?
Why don't we use something like  `Array.new(options["fields"]) | introspected_fields`

lib/generators/blueprinter/blueprint_generator.rb
```diff
@@ -98,7 +98,7 @@ def formatted_fields
 
       def associations
         as = if options["detect_associations"]
-               [].concat(options["associations"], introspected_associations.keys)
+               Array.new(options["associations"]).concat(introspected_associations.keys)
```
philipqnguyen: same concern here too regarding duplicate associations given.

lib/generators/blueprinter/initializer_generator.rb
```diff
@@ -0,0 +1,56 @@
...
+        {oj: "Oj", yajl: "Yajl::Encoder"}[options["generator"].intern] if options["generator"]
+      end
+
+      def empty_string(default)
+        default.to_s.gsub(/"'/,'').blank?
+      end
+
+    end
+  end
+end
```
philipqnguyen: Would you make a separate PR for the initializers generator? Reasoning is that I would like to have a more thorough discussion regarding generating initializers and the default configurations, and I don't want to have that block the rest of this PR. Currently, I think it's not very useful to generate an empty initializer by default.
```ruby
end
```

Rather I'd like to have the initializer generator to generate all the defaults configs of blueprinter, with code comments showing potential options.
Something like
```ruby
  # The default generator is JSON. You may replace it with OJ or Yajil as well.
  config.generator = JSON
end
```

spec/generators/blueprint_generator_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+    describe "given -i" do
+      include_examples "generated_file"
+      before do
+        run_generator %W(#{model} -i)
+      end
+
+      it "blueprint file has identifier" do
+        is_expected.to contain(/identifier/)
+      end
+    end
```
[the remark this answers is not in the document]
wlkrw: i've fixed and added a test

lib/generators/blueprinter/blueprint_generator.rb
wlkrw: the array gets ```uniq``` called on it; I added a test for this

lib/generators/blueprinter/initializer_generator.rb
wlkrw: sounds good. I took that part out. I'll wait for the next pr until this is ready to merge in case I drop off the internet again ;)

wlkrw:
@philipqnguyen I believe I've addressed all your change requests. Thanks for your patience.

lib/generators/blueprinter/blueprint_generator.rb
```diff
@@ -0,0 +1,127 @@
...
+        introspected_name = if introspected_associations[association_name].respond_to?(:klass)
+                              introspected_associations[association_name].klass.to_s
+                            else
+                              nil
+                            end
+        "#{introspected_name || association_name.camelcase}Blueprint"
+      end
+
+      def indent
+        user_intended = {two: "  ", four: "    ", tab:"\t"}[options["indentation"].intern]
```
mcclayton: Minor: should this be user_indented?

philipqnguyen: can be followed up after

PR #184, opened 2019-10-10
Ability to include multiple views in a single method call

narendranvelmurugan:
Created a new method called include_views and pass multiple views as arguments, which will render all the passed views in the blueprint.  

before:
```
include_view :user_summary
include_view :org_summary
```

Now:
```
include_views :user_summary,  :org_summary
```

### DIFF
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -336,11 +336,11 @@ module Blueprinter
     end
 
 
-    # Specify another view that should be mixed into the current view.
+    # Specify additional views that should be mixed into the current view.
     #
-    # @param view_name [Symbol] the view to mix into the current view.
+    #  @param view_name [Array<Symbol>] the views to mix into the current view.
     #
-    # @example Including a normal view into an extended view.
+    # @example Including the normal and special views into an extended view.
     #   class UserBlueprint < Blueprinter::Base
     #     # other code...
     #     view :normal do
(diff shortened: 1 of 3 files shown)


lib/blueprinter/base.rb
```diff
@@ -336,6 +336,34 @@ def self.include_view(view_name)
     end
 
 
+    # Specify another view that should be mixed into the current view.
```
mcclayton:
```suggestion
    # Specify additional views that should be mixed into the current view.
```

lib/blueprinter/base.rb
```diff
@@ -336,6 +336,34 @@ def self.include_view(view_name)
...
+    #
+    # @param view_name [Symbol] the view to mix into the current view.
+    #
+    # @example Including a normal view into an extended view.
```
mcclayton:
```suggestion
    # @example Including the normal and special views into an extended view.
```

lib/blueprinter/base.rb
```diff
@@ -336,6 +336,34 @@ def self.include_view(view_name)
     end
 
 
+    # Specify another view that should be mixed into the current view.
+    #
+    # @param view_name [Symbol] the view to mix into the current view.
```
mcclayton:
```suggestion
    # @param view_name [Array<Symbol>] the views to mix into the current view.
```

mcclayton:
Thanks so much for contributing!

PR #185, opened 2019-10-14
Sorting Fields in Blueprinter

narendranvelmurugan:
```
class TestBlueprint < Blueprinter::Base
  fields :id, :state, :catalog
  view :expanded do
    association :task, blueprint: TaskBlueprint
  end
end
```
Actual Output:

```
{:task=>{:description =>"task assigned"}, :id=>1, :state=>1, :catalog=>nil} 
```

Expected Output:
```
{:id=>77, :state=>1, :catalog=>nil, :task=>{:description=>"task assigned"}} 
```
This resolves <a href =https://github.com/procore/blueprinter/issues/172> 172 </a>

### DIFF
--- a/lib/blueprinter/view_collection.rb
+++ b/lib/blueprinter/view_collection.rb
@@ -59,7 +59,7 @@ module Blueprinter
     end
 
     def merge_fields(source_fields, included_fields)
-      if sort_by_definition
+      if sort_by_definition && source_fields.blank?
         included_fields.merge(source_fields)
       else
         source_fields.merge(included_fields)


philipqnguyen:
As far as I can tell, this change just moves fields in `view` to the end of the output. What if the `view` is declared in the middle of the blueprint?

assuming this blueprint:
```ruby
class TestBlueprint < Blueprinter::Base
  field :should_be_first
  view :expanded do
    field :should_be_second
  end
  field :should_be_last
end
```

and given `TestBlueprint.render(obj, view: :expanded)`

The output in this change is:
```json
{
  "should_be_first": "first",
  "should_be_last": "last",
  "should_be_middle": "middle"
}
```

When the expected output should have fields ordered as defined:
```json
{
  "should_be_first": "first",
  "should_be_middle": "middle",
  "should_be_last": "last"
}
```

PR #189, opened 2019-11-14
Honor default option for has_many associations

GWillmann:
Has many associations return an empty relation instead of a nil value so the default option was not honored:

`return default_value(options) if value.nil?`

I added another condition to it:

`return default_value(options) if value.nil? || (value.blank? && value.is_a?(ActiveRecord::Associations::CollectionProxy))`

### DIFF
--- a/spec/integrations/base_spec.rb
+++ b/spec/integrations/base_spec.rb
@@ -236,11 +236,6 @@ describe '::Base' do
       end
 
       context "Given association is empty" do
-        before do
-          empty_relation = obj.vehicles
-          expect(obj).to receive(:vehicles).and_return(empty_relation)
-        end
-  
         context "Given default association value is provided and is empty" do
           let(:result) { '{"id":' + obj_id + ',"vehicles":null}' }
           
--- a/spec/integrations/base_spec.rb
+++ b/spec/integrations/base_spec.rb
@@ -234,6 +234,31 @@ describe '::Base' do
           end
         end
       end
+
+      context "Given association is empty" do
+        before do
+          empty_relation = obj.vehicles
+          expect(obj).to receive(:vehicles).and_return(empty_relation)
+        end
+  
+        context "Given default association value is provided and is empty" do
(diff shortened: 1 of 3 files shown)


spec/integrations/base_spec.rb
```diff
@@ -234,6 +234,31 @@ def extract(association_name, object, _local_options, _options={})
           end
         end
       end
+
+      context "Given association is empty" do
+        before do
+          empty_relation = obj.vehicles
+          expect(obj).to receive(:vehicles).and_return(empty_relation)
```
AllPurposeName: I think this assertion is unnecessary. We should know the state of our testing environment at this point.

If this expectation failed, the later expectations would also fail, correct?

GWillmann: You're right, I have removed the useless assertion!

mcclayton:
Hey @GWillmann, first off -- thank you so much for your contribution! 
@philipqnguyen was a little quick to add the `bug` label, as this is really a breaking change and a more of a feature request than a bug. 
The docs currently state that default options are specifically only applied to `nil` field/association evaluations.
Philip and I did, however, discuss some non-breaking API changes we could make to support this and similar use cases that have come up. 
Would the ability to provide `nil-able` values that essentially get cast/converted to `nil` solve your use case?

i.e.
```rb
  identifier :uuid

  view :normal do
    # We want "" to be cast to nil, by additionally specifying a default, "" values become "N/A"
    field :first_name, default: "N/A", nillable: Blueprinter::EmptyString
    # We want empty collections to be cast to nil
    association :projects, blueprint: ProjectBlueprint, nillable: Blueprinter::EmptyCollection
  end
end
```

GWillmann:
That should indeed help my use case, thanks! Closing it in favor of https://github.com/procore/blueprinter/pull/191

philipqnguyen:
yup! I mislabled it a bug, but this is indeed would be a seperate feature. thanks @mcclayton for clearing that up.

PR #191, opened 2019-12-19
Add field/association level `default_if` option

mcclayton:
Sometimes, you may want certain "empty" values to pass through to the default value other than nil.
This PR gives Blueprinter the ability to treat the following empty types as the default value (or `nil` if no default provided).

#### Blueprinter::EMPTY_COLLECTION
An empty array or empty active record collection.

#### Blueprinter::EMPTY_HASH
An empty hash.

#### Blueprinter::EMPTY_STRING
An empty string or symbol.

#### Field-level/Association-level Setting
```ruby
  identifier :uuid

  view :normal do
    # If first_name is an empty string, it will become "N/A"
    field :first_name, default_if: Blueprinter::EmptyString, default: "N/A"
    # If the projects association collection is empty, it will become nil
    association :projects, blueprint: ProjectBlueprint, default_if: Blueprinter::EmptyCollection
  end
end
```


lib/blueprinter/empty_types.rb
```diff
@@ -0,0 +1,25 @@
+require_relative 'helpers/type_helpers'
+
+module Blueprinter
+  EMPTY_COLLECTION = "empty_collection".freeze
+  EMPTY_HASH = "empty_hash".freeze
+  EMPTY_STRING = "empty_string".freeze
+
+  module EmptyTypeHelper
```
wai-procore: wondering if this module name should be consistent with the file name `empty_types.rb` or vice versa?

lib/blueprinter/version.rb
```diff
@@ -1,3 +1,3 @@
 module Blueprinter
-  VERSION = '0.20.0'.freeze
+  VERSION = '0.20.1'.freeze
```
philipqnguyen: This is new feature. What do you think if we make it 0.21.0?

Also need update to CHANGELOG

mcclayton: Of course, great catch!

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.21.0  - 2019/12/19
+* 🚀 [FEATURE] Ability to specify `default_if` field/association option for more control on when the default value is applied. [191](https://github.com/procore/blueprinter/pull/191). Thanks to [@mcclayton](https://github.com/mcclayton).
```
philipqnguyen: the vanity =D

mcclayton: Just following the template 😅

PR #193, opened 2019-12-23
Add before_render callbacks

peterzhu2118:
Adds `before_render` callbacks before the blueprint is serialized. My use case is to load a database record relevant to the blueprint when rendered.

The syntax looks like the following:
```ruby
  before_render do |object, options|
    options[:address] = "REDACTED" if options[:redact]
  end
  identifier :uuid
  field :address { |object, options| options[:address] || object.address }
end
```

In this example, address is set to the string `REDACTED` if the `redact` field is passed into `options`.

It also supports `before_render` callbacks nested in `view`s:
```ruby
  before_render do |object, options|
    options[:address] = object.address.country if options[:redact]
  end
  identifier :uuid
  field :address { |object, options| options[:address] || object.address }

  view :medium_privacy do
    before_render do |object, options|
      options[:address] = object.address.city + ", " + options[:address]
    end
  end
end
```

In which case, the call `UserBlueprint.render(user, view: :medium_privacy, redact: true)` output something like this:
```ruby
{
  "address": "Ottawa, Canada"
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -697,6 +697,47 @@ _NOTE:_ The field-level setting overrides the global config setting (for the fie
 
 
 <details>
+<summary>before_render callbacks</summary>
+
+---
+
+Callback called before fields are serialized. This can be used to compute
+values used in your fields and storing computed valued inside the
+options hash.
+
+#### Example
+
+```ruby
+class UserBlueprint < Blueprinter::Base
+  before_render do |object, options|
+    options[:address] = "REDACTED" if options[:redact]
+  end
+
+  identifier :uuid
+  field :address { |object, options| options[:address] || object.address }
+end
+```
+
+Useage:
+
+```ruby
+puts UserBlueprint.render(user, view: :extended, redact: true)
+```
+
+Output:
+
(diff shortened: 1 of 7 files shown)


philipqnguyen:
@peterzhu2118  I really like the name `before_render`, but aside from that, does this achieve anything that the current `transform` method is not able to achieve?

peterzhu2118:
Hey @philipqnguyen, I think `transform` runs after the blueprint fields have been evaluated. However, `before_render` runs before the blueprint fields are evaluated. My use case is to load a database record in `before_render` based on the record passed in and the current user passed into the options. This database record is then used to populate various fields in the blueprint.

heebj:
This feature would be really helpful for me as well. Are there any news on this? Thanks a lot!

adamwells:
We've reviewed this in our maintainers meeting and determined that this behavior falls outside the scope of our vision for blueprinter. For now we are keeping this kind of object preparation outside of blueprinter, which should be responsible for object serialization only

Issue #178, opened 2019-08-29
Configurable Default Extractor?


philipbjorge:
We're interested in potentially extending the `AutoExtractor` for our usecases and rather than monkey patching are wondering if you'd consider taking a change that makes the default extractor configurable?

https://github.com/procore/blueprinter/issues/177

philipqnguyen:
@philipbjorge sorry for the late reply. Yes I definitely like this idea. Specifically setting a default extractor in the `configuration` would be a good way to go. If not set, it would be the `AutoExtractor`. If you have the time to make a PR for this, we would definitely welcome it.

Apologies for the late response.

mcclayton:
Closing since this has been implemented in: https://github.com/procore/blueprinter/pull/198

PR #198, opened 2020-01-29
Configure extractor

wlkrw:
This is a fix for #178.

Since I've never personally needed this, I'm afraid the example added to README.md does not do a great job motivating why to use this feature. I'm open to suggestions or edits.


philipqnguyen:
released in 0.23.0

PR #201, opened 2020-02-10
Fixes include_view order to make field/association redefinition works

### LINKED ISSUE #200: Issue with include_view order and field/association redefinition
Hello!

We have a case where we use the same association several times in a Serializer, with different views. Because of order in which `include_view` works, the specified view is sometimes not respected:

```ruby
class Serializer
  association :children

  view :extended do
    association :children, view: :extended # This works
  end

  view :community do
    include_view :extended
    association :children, view: :community # This will render children with the extended view and not the community
  end
end
```

I'm open to make a PR to change the inclusion order. I wanted to check with you beforehand, because this may be a breaking change since it will change the redefinition order.

What do you think?

Berardpi:
Fixes #200

### DIFF
--- a/lib/blueprinter/view_collection.rb
+++ b/lib/blueprinter/view_collection.rb
@@ -48,7 +48,6 @@ module Blueprinter
     def sortable_fields(view_name)
       excluded_fields = {}
       fields = views[:default].fields
-      fields = merge_fields(fields, views[view_name].fields)
       views[view_name].included_view_names.each do |included_view_name|
         next if view_name == included_view_name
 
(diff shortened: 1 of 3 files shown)


kayakyakr:
Been using blueprinter for all of 1 day, and already wishing that this was merged.

CalvinWalzel:
I would love for this to be merged :)

mcclayton:
Firstly, I just want to apologize for how long it took to get this merged -- and secondly, I want to say a huge thank you for fixing this issue @Berardpi! I'll get a new release drafted up with this fix shortly.

PR #214, opened 2020-05-04
[Extractor] Add delegate extractor

ThanhKhoaIT:
Add delegate extractor to instead the simple block

With simple block
```ruby
class ProjectBlueprint < Blueprinter::Base
  identifier :uuid
  field :company_name do |project, _options|
    project.user.company
  end
end
```
and we can use
```ruby
class ProjectBlueprint < Blueprinter::Base
  identifier :uuid
  field :company_name, delegate: { to: :user, source: :company }
end
```

Usage:
```ruby
puts ProjectBlueprint.render(project)
```
Output:
```json
{
  "company_name": "My Company LLC"
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,6 +306,49 @@ end
 ---
 </details>
 
+<details>
+<summary>Delegate</summary>
+
+---
+
+You can use the delegate instead of the block.
+
+With Block:
+```ruby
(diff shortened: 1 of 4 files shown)


aka47:
Why did you close this PR - @ThanhKhoaIT ? Looks what I was looking for :)

PR #217, opened 2020-05-12
Allow pasing :method option to `field`

ghost:
This change allows us to write our blueprints in terms of the JSON structure that we expect:

Before:
```ruby
class MyBlueprint < Blueprint::Base
  identifier :id
  field :name
  field :item_number, name: :position
end
```

After:
```ruby
class MyBlueprint < Blueprint::Base
  identifier :id
  field :name
  field :position, method: :item_number
end
```

Both blueprints would output
```json
{
  id: 1,
  name: "A name",
  position: 123
}
```

I think the latter method seems preferable and more readable as it reflects what keys are supposed to show up in the JSON result.

This change should not break existing blueprint structures; it just allows more flexibility for those who prefer to write their blueprints the way that they expect the JSON object should appear.

### DIFF
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -57,23 +57,28 @@ module Blueprinter
         name,
         extractor,
         self,
-        block: block,
+        block: block
       )
     end
 
     # Specify a field or method name to be included for serialization.
     # Takes a required method and an option.
     #
-    # @param method [Symbol] the field or method name you want to include for
-    #   serialization.
+    # @param key [Symbol] the field or name you want to include for
+    #   serialization. If :method is not specified as an option, this param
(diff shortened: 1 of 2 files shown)


AllPurposeName:
This is much more intuitive for me. Another point for it is when you switch a name field around to a block defined field it is easier.
```ruby
# Replacing
# field :item_number, name: :position
field :position do |my_bloop|
  my_bloop.try(:item_number) || my_bloop[:item_number]
end
```

```ruby
# Replacing
# field :position, method: :item_number
field :position do |my_bloop|
  my_bloop.try(:item_number) || my_bloop[:item_number]
end
```

That one always gets me.

lessthanjacob:
Closing this as the original author's account has been deleted.

PR #219, opened 2020-05-27
Fix Changelog timestamps

leonhess:
The Year for all 0.23.* releases but the 0.23.4 release is 2019 but it should be 2020.


mcclayton:
Thanks for catching 😄

PR #220, opened 2020-06-23
Enable passing options between blueprints via associations.

mcclayton:
Enable passing options from one Blueprint to another via an association.

For example:
```ruby
class VehicleBlueprint < Blueprinter::Base
  identifier :uuid
  field :full_name do |vehicle, options|
    "#{vehicle.model} #{options[:trim]}"
  end
end
class DriverBlueprint < Blueprinter::Base
  identifier :uuid
  view :normal do
    association :vehicles, blueprint: vehicle_blueprint, options: { trim: 'LX' }
  end
end
```

### DIFF
--- a/lib/blueprinter/field.rb
+++ b/lib/blueprinter/field.rb
@@ -10,6 +10,9 @@ class Blueprinter::Field
   end
 
   def extract(object, local_options)
+    # If this is an association field, merge the association field options in
+    # to the render options
+    local_options = local_options.merge(options[:options]) if association? && options[:options]
     extractor.extract(method, object, local_options, options)
   end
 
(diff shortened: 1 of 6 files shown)


lib/blueprinter/field.rb
```diff
@@ -10,6 +10,9 @@ def initialize(method, name, extractor, blueprint, options = {})
   end
 
   def extract(object, local_options)
+    # If this is an association field, merge the association field options in
+    # to the render options
+    local_options = local_options.merge(options[:options]) if association? && options[:options]
```
philipqnguyen, line 15: Why don't we put this in the `AssociationExtractor`, since this is related to associations?

lib/blueprinter/version.rb
```diff
@@ -1,3 +1,3 @@
 module Blueprinter
-  VERSION = '0.23.4'.freeze
+  VERSION = '0.23.5'.freeze
```
philipqnguyen, line 2: should this be 0.24.0?

As this is a new feature.

PR #233, opened 2020-08-08
Raise errors for invalid blueprints for associations

caws:
This PR validates that a given Blueprint class passed for an association is valid and inherits from `Blueprinter::Base`.

**The issue.**

Take the following `Blueprinter`.

```ruby
class PaymentBlueprinter < Blueprinter::Base
    identifier :id
    view :estimate do
      association :client_taxes, blueprint: Client, view: :estimate
     # or 
     # association :client_taxes, blueprint: {}, view: :estimate
    end
end
```

This raises ```undefined method `prepare' for #<Class:0x00005649611cdb70> Did you mean? prepend``` when [this line](https://github.com/procore/blueprinter/blob/eb46aef170b29a3df29d741fc249713382cff0c4/lib/blueprinter/extractors/association_extractor.rb#L18) is interpreted.

**The fix:**

Add logic to raise errors if `Blueprinter::Base` is not present in the ancestor list of the class provided as the Blueprinter for a given association.

My thoughts:

The undefined method error is not intuitive and does offer any hint as to what's going on to the programmer. As an extra I also added a check to see if the provided Blueprint class has any ancestors at all.


lib/blueprinter/extractors/association_extractor.rb
```diff
@@ -24,8 +24,36 @@ def default_value(association_options)
...
+      return if blueprint.ancestors.include? Blueprinter::Base
+
+      # Raise error describing what's wrong.
+      raise BlueprinterError, "Class #{blueprint.name} does not inherit from "\
+                              'Blueprinter::Base and is not a valid Blueprinter '\
+                              "for #{association_name} association."
     end
   end
-end
+end
```
philipqnguyen, line 59: missing additional line

lib/blueprinter/extractors/association_extractor.rb
```diff
@@ -7,14 +7,14 @@ def initialize
...
-      options_without_default = options.reject { |k,_| k == :default || k == :default_if }
+    def extract(association_name, object, local_options, options = {})
+      options_without_default = options.reject { |k, _| k == :default || k == :default_if }
       # Merge in assocation options hash
       local_options = local_options.merge(options[:options]) if options[:options].is_a?(Hash)
       value = @extractor.extract(association_name, object, local_options, options_without_default)
       return default_value(options) if use_default_value?(value, options[:default_if])
       view = options[:view] || :default
-      blueprint = association_blueprint(options[:blueprint], value)
+      blueprint = association_blueprint(options[:blueprint], value, association_name)
```
philipqnguyen, line 17: Instead of performing the blueprint validation checks here, would you instead perform the validation checks here in `association` https://github.com/procore/blueprinter/blame/eb46aef170b29a3df29d741fc249713382cff0c4/lib/blueprinter/base.rb#L159 instead?

The reasoning is that there is already a check there, but it's not as thorough as what you have here. Let's add your checks there and validate it all in one place. Additionally, the `Base.association`  is happening during initialize time, where as the operation here is happening during runtime. During initialize time, blueprinter is mapping all the blueprints and field names, while during runtime it extracts field values. It's better to do the class validation during initialize time since we want to limit operations in runtime for performance reason.

caws: Sure, makes perfect sense. I hadn't noticed that there were some validations in place already. I did what you suggested and also moved [this validation](https://github.com/procore/blueprinter/blame/eb46aef170b29a3df29d741fc249713382cff0c4/lib/blueprinter/base.rb#L159) into the `base_helper.rb`.

lib/blueprinter/helpers/base_helpers.rb
```diff
@@ -62,6 +63,44 @@ def validate_root_and_meta(root, meta)
         end
       end
 
+      def dynamic_blueprint?(blueprint)
+        blueprint.is_a?(Proc)
+      end
+
+      def validate_blueprint(blueprint, method)
```
mcclayton, line 70: Would you please add the bang to each of these validation methods that possibly throw errors
```suggestion
      def validate_blueprint!(blueprint, method)
```

caws: Yes. I had initially not added the bang to the end of the method names because I noticed that [`validate_root_and_meta`](https://github.com/procore/blueprinter/blob/eb46aef170b29a3df29d741fc249713382cff0c4/lib/blueprinter/helpers/base_helpers.rb#L54) didn't have it.  Should be good now.

Issue #160, opened 2019-06-20
Eager loading problem


denmarkmeralpis:
I dunno if this issue is valid but here's the definitions I have:

```ruby
# order_group.rb
class OrderGroup < ApplicationRecord
  has_many :orders
end

# order.rb
class Order < ApplicationRecord
  belongs_to :order_group
  has_many :order_lines
  has_many :fees
end

# blueprints

# order_group_blueprint.rb
view :extended do
  association :orders, blueprint: OrderBlueprint, view: :extended
end

# order_blueprint.rb
view :extended do
  association :order_lines, blueprint: OrderLineBlueprint
  association :fees, blueprint: FeeBlueprint
end

# some_controller.rb
...
OrderGroupBlueprint.render_as_hash(OrderManifest.all, view: :extended)
...
```

Now, how do I eager load the `fees` & `order_lines` when using the blueprint `OrderGroupBlueprint`? 

Feel free to close the issue if this is invalid :) Thank you very much

denmarkmeralpis:
Sorry, this is a duplicate issue. 

Referencing this to #96 

Closing this now. Cheers!

denmarkmeralpis:
Finally got a solution

```ruby
field :orders do |b|
  b.orders.includes(:fees, :order_lines)
end
```

Closing this issue now. Thanks

tcd:
@denmarkmeralpis Could you elaborate on your solution?

tferrerITBA:
This isn't a great workaround, as you cannot specify which blueprint to use for that specific field. There has to be a better way.

philipqnguyen:
I'm trying to understand the problem. Is there a reason that you cannot eager load the associations before passing the `ActiveRecord::Relation` object to `SomeBlueprint.render(args)`?

PR #234, opened 2020-08-11
Association extractor with includes option

ThanhKhoaIT:
#160 

### Old solution
```ruby
# order_group_blueprint.rb
view :extended do
  association :orders, blueprint: OrderBlueprint, view: :extended do |order_group|
    order_group.orders.includes(:fees, :order_lines)
  end
end
```

### New solution
```ruby
# order_group_blueprint.rb
view :extended do
  association :orders, blueprint: OrderBlueprint, view: :extended, includes: [:fees, :order_lines]
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,6 +375,32 @@ class DriverBlueprint < Blueprinter::Base
 end
 ```
 
+Support eager loading to fix N+1 queries for the association.
+For example:
+```ruby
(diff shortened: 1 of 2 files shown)


maboelnour:
@ThanhKhoaIT why this is closed?

saxxi:
Why is this closed @ThanhKhoaIT ?

Issue #225, opened 2020-07-13
Not able to use transformer when it is in included view


bhooshiek-narendiran:
```
view :skeleton do
    include_view :summary
    excludes :actors, :ticket_id
  end

  view :summary do
    field :actors_response_info, name: :actors
    transform FieldTransformer
  end

```
Here when I directly use the view: summary then the FieldTransformer is working. But I can't able to use the same transformer when it is included in another view. Like if view: skeleton is called, the FieldTransformer is not working. This is because we are not taking the included views transformer while rendering the output. 
Can I raise a pr for this one?

adamwells:
@bhooshiek-narendiran Good find, thank you - definitely feel free to open a PR to solve this.

bhooshiek-narendiran:
Thanks, @adamwells Will do.

aarona:
Could really use this. Would be great if someone could do a code review and accept this PR. @bhooshiek-narendiran, thanks for fixing this btw.

cooperka:
Thanks for the PR @bhooshiek-narendiran!

Separate but related issue: views should inherit transformers from the base class but they do not. This is demonstrated by a new PR https://github.com/procore/blueprinter/pull/247. It is unfortunately **not** fixed by your change.

I wonder if we could make a generic fix for both issues... I don't currently have time to dig deeper but wanted to at least share these preliminary findings here (a workaround for me is to simply re-include the transformer in every single view, so that's what I'm doing until I have time to help with this proper fix).

bhooshiek-narendiran:
HI @cooperka! Can you elaborate me the above issue with an example? So I'll try to fix that one in this PR.

cooperka:
Yes please see the linked PR #247 with a working unit test (it's failing, in order to demonstrate the issue) 🙂 Let me know if you have questions related to it, thanks for your help!

bhooshiek-narendiran:
Hi, @cooperka One doubt whether the view transformer should override the class transformers(parent and child) or both should execute (view transformers + class transformers)?

cooperka:
Hmm, if both execute, is there a way to STOP the parent from executing through some syntax? If there's no way to stop it, I think it must override entirely. But I will leave that up to the library author during PR review.

bhooshiek-narendiran:
[...]

PR #236, opened 2020-08-20
Allowing to use included views transformer from the current view

bhooshiek-narendiran:
Fix for [#225](https://github.com/procore/blueprinter/issues/225)


joaovitoras (approved):
🎉

lib/blueprinter/view_collection.rb
```diff
@@ -29,8 +29,10 @@ def fields_for(view_name)
       (identifier_fields + sorted_fields).reject { |field| excluded_fields.include?(field.name) }
     end
 
-    def transformers(view_name)
-      views[view_name].transformers
+    def transformers_for(view_name)
+      view_transformers = Set.new(transformers(view_name))
```
njbbaer, line 33: You'll need `require 'set'` for this to work.

njbbaer (changes requested):

lib/blueprinter/view_collection.rb
```diff
@@ -29,8 +29,10 @@ def fields_for(view_name)
...
+      views[view_name].included_view_names.each { |included_view_name| view_transformers |= transformers(included_view_name) }
```
njbbaer, line 34: This only works for directly included views, but not nested included views.
```
view :normal do
  transform TestTransformer
  fields :first_name
end

view :extended do
  include_view :normal
end

view :all do
  include_view :extended
end
```

lib/blueprinter/view_collection.rb
njbbaer, line 33: We haven't used `Set` in the project for anything else. Does it make sense just for this?

lib/blueprinter/view_collection.rb
```diff
@@ -29,8 +29,10 @@ def fields_for(view_name)
       (identifier_fields + sorted_fields).reject { |field| excluded_fields.include?(field.name) }
     end
 
-    def transformers(view_name)
-      views[view_name].transformers
+    def transformers_for(view_name)
```
njbbaer, line 32: The meaning of `transformers_for` is unclear to me. Shouldn't this remain `transformers`?

lib/blueprinter/view.rb
```diff
@@ -14,10 +14,6 @@ def initialize(name, fields: {}, included_view_names: [], excluded_view_names: [
       @sort_by_definition = Blueprinter.configuration.sort_fields_by.eql?(:definition)
     end
 
-    def transformers
-      view_transformers.empty? ? Blueprinter.configuration.default_transformers : view_transformers
-    end
```
njbbaer, line 19: Why remove this rather than update? Was it unused?

njbbaer:
I fixed the conflicts on this branch. @bhooshiek-narendiran please reply if you're interested in getting this merged. I think it would be a beneficial addition.

bhooshiek-narendiran:
@njbbaer Yes we can work on this.

lib/blueprinter/view_collection.rb
bhooshiek-narendiran: Done

lib/blueprinter/view_collection.rb
bhooshiek-narendiran: Removed

lib/blueprinter/view_collection.rb
bhooshiek-narendiran: Retained as it is

lib/blueprinter/view.rb
bhooshiek-narendiran: Retained as it is.

bhooshiek-narendiran:
@njbbaer Can you check this once? Will push the specs parallely.

lib/blueprinter/view_collection.rb
```diff
@@ -35,7 +35,16 @@ def fields_for(view_name)
...
+      transformers = gather_transformers_from_included_views(view_name)
+      transformers.presence || views[:default].transformers
+    end
+
+    def gather_transformers_from_included_views(view_name)
+      current_view = views[view_name]
+      current_view.included_view_names.each_with_object([]) do |included_view_name, transformers|
+        next if view_name == included_view_name
+        transformers.concat(gather_transformers_from_included_views(included_view_name))
+      end.concat(current_view.view_transformers)
```
ritikesh, line 47: I think we need to ensure that transformers are run "at-most" once. we can do that by using `.uniq` or by using a `Set` instead of `Array` in `each_with_object`. Does this also solve the inheritance [problem](https://github.com/procore-oss/blueprinter/issues/225)? IMO, we need to ensure transformers get inherited from base class as well.

bhooshiek-narendiran: Sure @ritikesh. Will add Uniq. And yes it will take the included views transformers in the included order. 
If no transformers are present, it will take the default views transformer or else global transformer

bhooshiek-narendiran:
Moved to this [363](https://github.com/procore-oss/blueprinter/pull/363)

PR #250, opened 2020-11-25
Update README.md for Rails Controller usage

dominikdarnel:


README.md
```diff
@@ -49,6 +49,26 @@ And the output would look like:
...
+---
+
+You can use your blueprint in a Rails API controller to render yout JSON:
+
+```ruby
+class UsersController < ActionController::API
+  ...
+  def show
+    user = UserBlueprint.render_as_hash(@user)
+    render json: user
```
swiknaba, line 64: no need to render as hash, just do

```ruby
def show
    render json: UserBlueprint.render(@user)
end
```

naming wise, I would choose with `UserSerializer` --after all, the file is a serializer. You are just using blueprint**er** as a backend for the serializer; you could as well swap that for any other gem.

README.md
```diff
@@ -49,6 +49,26 @@ And the output would look like:
 ---
 </details>
 
+<details open>
+<summary>Rails Controller</summary>
+
+---
+
+You can use your blueprint in a Rails API controller to render yout JSON:
```
swiknaba, line 57: `use your blueprint` --> `use blueprinter
`render yout JSON` --> `render an ActiveRecord object`

PR #254, opened 2021-01-21
Raising BlueprinterError on some situations

wollistik:
* Raising errors on invalid included views
* ReRaising error with details on extraction error

### DIFF
--- a/lib/blueprinter/extractor.rb
+++ b/lib/blueprinter/extractor.rb
@@ -2,12 +2,16 @@
 
 module Blueprinter
   class Extractor
+    FieldExtractionError = Class.new(BlueprinterError)
+
     def extract(_field_name, _object, _local_options, _options = {})
       raise NotImplementedError, 'An Extractor must implement #extract'
     end
 
     def self.extract(field_name, object, local_options, options = {})
       new.extract(field_name, object, local_options, options)
+    rescue StandardError
+      raise Blueprinter::Extractor::FieldExtractionError, "Error when extracting value for field: '#{field_name}'"
     end
   end
 end
--- a/lib/blueprinter/formatters/date_time_formatter.rb
+++ b/lib/blueprinter/formatters/date_time_formatter.rb
@@ -3,6 +3,7 @@
 module Blueprinter
   class DateTimeFormatter
     InvalidDateTimeFormatterError = Class.new(BlueprinterError)
+    DateTimeFormattingError = Class.new(BlueprinterError)
 
     def format(value, options)
       return value if value.nil?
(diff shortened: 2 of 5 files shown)


mcclayton:
This is valuable feedback @wollistik, and we appreciate the PR! The core maintainers will take a look at this PR soon and provide feedback in carrying that through. Thanks!

lib/blueprinter/helpers/base_helpers.rb
```diff
@@ -44,7 +44,15 @@ def inherited(subclass)
       def object_to_hash(object, view_name:, local_options:)
         result_hash = view_collection.fields_for(view_name).each_with_object({}) do |field, hash|
           next if field.skip?(field.name, object, local_options)
-          hash[field.name] = field.extract(object, local_options)
+
+          begin
```
kriansa, line 48: As this is a rather non-standard practice in popular codebases, I would also add a comment here stating that the purpose of this is to wrap all errors under a `BlueprinterError` so it makes it clearer for people non-familiar with that kind of pattern.

kriansa (approved):
I really appreciate libraries that wrap errors under a singular class so I can safely `rescue` and handle them without any weird matching and workarounds.

:heart_decoration: :bow:

lessthanjacob:
Hey @wollistik! Apologies for the delay here. 

Looking at the proposed changes, I do appreciate the additional of more intentional error handling! However, I'd actually advocate more towards implementing error classes that describe the error via their type, as opposed to needing to introspect the included message of a generic error class.

We should be able to get the best of both worlds here by simply subclassing each error class from `BlueprinterError`: Users can explicitly handle a `Blueprinter::FieldExtractionError`, _or_ they could rescue `BlueprinterError` generally.

wollistik:
Hi @lessthanjacob ,

a lot of time has passed since the initial PR :wink:  I now also prefer designated error classes and adapted the code and the test cases accordingly. I took a sample from the `DateFormatter` class and put the error class definitions where they are used. But if you prefer them to be all at one file (e.g. directly side by side with `BluprinterError`), I can adjust my PR.

lessthanjacob (commented):
> a lot of time has passed since the initial PR 😉 I now also prefer designated error classes and adapted the code and the test cases accordingly. 

I appreciate your patience here, and for coming back and continuing to work on this after so much time as passed! And apologies for the delay here again. 

Looking through this PR again, I'd suggest avoiding _broadly_ rescuing from `StandardError` unless we view it as necessary. There's definitely some benefit in doing so, but I'd be concerned that we'd be obscuring bugs that are originating from  "integrating" code behind a `BlueprinterError`. 

However, I'd be happy to discuss this further! I'd be interested in getting perspective from the rest of the maintainers here as well (@procore-oss/procore-blueprinter).

jmeridth:
Not stale /cc @njbbaer

Issue #228, opened 2020-07-20
Calling missing view


CarlosUvaSilva:
is there a way to instead of return `View 'private' is not defined>` to just return the "normal"/default view? 
my case is rendering a polymorphic association like 
```
association :owner, blueprint: ->(taskable) {
    taskable.blueprint
  }, view: :private, default: {}
```

imagine that Owner can be either a User or an Admin. I want the user's private view but admin doesn't have a private so I'd like to just kind of ignore it
Current fix is to add 
```
view :private do
  end
``` 
to my admin blueprint :man_shrugging: which is really annoying because I have way mores cases than this one. My app is polymorphic galore

EDIT

I've since monkey patched with
```
module Blueprinter
  class Base
    def self.prepare(object, view_name:, local_options:, root: nil, meta: nil)
      unless view_collection.has_view? view_name
        # raise BlueprinterError, "View '#{view_name}' is not defined"
        view_name = :default
      end

      data = prepare_data(object, view_name, local_options)
      prepend_root_and_meta(data, root, meta)
    end
  end
end
```
but I'm not sure if I'm happy with this

PR #263, opened 2021-03-14
feature: render default view instead of raising error if view is missing

CarlosUvaSilva:
I was having trouble with missing views while using polymorphic relationships and transformers.
I've added this new configuration to prevent an error being raised and instead using the default view if a view is not defined

this is related to my issue #228

### DIFF
--- a/README.md
+++ b/README.md
@@ -868,6 +868,44 @@ Output:
 ---
 </details>
 
+<details>
+<summary>Missing Views</summary>
+
+---
+
+By default a BlueprinterError is raised if you call a view that doesn't exist. This is troublesome in some cases while using polymorphic relationships.
+If you set this config to `:default` it will render the default view instead of raising an error.
+
+Usage:
+
+```ruby
+Blueprinter.configure do |config|
+  config.missing_view = :default
+end
+```
+
+```ruby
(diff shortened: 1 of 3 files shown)


mcclayton:
Hi @CarlosUvaSilva, thank you so much for contributing to Blueprinter!
I'm curious, instead of handling this with a fallback configuration, would having access to the list of view names help you to programmatically handle this case?
https://github.com/procore/blueprinter/pull/256

CarlosUvaSilva:
Hello @mcclayton 

I don't that would help that much. I mean my proposal is a simple configuration because I understand and agree that raising an error should be the default, I use it like that in several projects.

The way I'd use PR #256 would be something like `Blueprinter.view_names.include?(my_view)` in my "Json Response Builder".
I'd rather use a simple configuration and not more code inside my project, that's why I was monkeypatching in my original issue #228

dlcarter:
Thank you for contributing. There are a number of other related PRs and concerns we are taking into consideration. We are looking into how to best consolidate these approaches into one implementation. This effort can be tracked with https://github.com/procore/blueprinter/projects/2#card-57792153

CarlosUvaSilva:
sorry to grave dig this but any news?
starting a new project and using my own fork again

PR #279, opened 2021-09-07
Add empty JSON as a valid empty type

ramongr:
Adds a new empty type called `Blueprinter::EMPTY_JSONB` to represent possible empty values from jsonb columns in databases.


philipqnguyen:
Correct me if I'm wrong, the most likely scenario for using JSONB is when there is a JSONB column in a postgresql column. However, when you access that corresponding attribute in Rails, it would already be deserialized into a Hash right? So I don't see why we need a JSONB type here.

PR #294, opened 2022-07-26
Dy/pm 15994 query optimization

davidyeeprocore1:
Currently, blueprinter does not attempt to eager load associations for collections of activerecords (ActiveRecordRelation). This results in n+1 queries being performed. The way this is resolved at the moment is the client (application that ingests blueprinter) must perform preloading. This PR aims to solve the n+1 query issue from within blueprinter rather than placing the burden on the client.


lib/blueprinter/helpers/base_helpers.rb
```diff
@@ -6,6 +6,8 @@ def self.included(base)
 
     module SingletonMethods
       include TypeHelpers
+      include AssociationHelpers
+      @@should_preload_associations = false
```
philipqnguyen, line 10: Two things: 
1. Instead of `should_preload_associations`, it should just be `preload_associations`. Otherwise it seems unsure and somewhat like a test file.
2. This should be in configurations.rb https://github.com/procore/blueprinter/blob/60b8f94616e18aa0346fa3c697b9e30c66c4ec8c/lib/blueprinter/configuration.rb#L3-L18 so that users can configure their options like everything else.

lib/blueprinter/helpers/base_helpers.rb
```diff
@@ -17,12 +19,28 @@ def prepare_for_render(object, options)
...
+        end
+      end
+
       def prepare_data(object, view_name, local_options)
         if array_like?(object)
-          object.map do |obj|
-            object_to_hash(obj,
-                           view_name: view_name,
-                           local_options: local_options)
+          if object.respond_to?(:reflect_on_all_associations) && preload_associations?
```
philipqnguyen, line 36: What happens if the object given is already eager loaded?

lib/blueprinter/helpers/base_helpers.rb
davidyeeprocore1: I see what you are saying here @philipqnguyen . One thing to note is that the `preload_associations` variable you are referencing is not meant to be a permanent field. It would only exist until we are done testing this feature. After that time, blueprinter will try to eager load associations that would otherwise not be eager loaded. My main concerns are: 

- I will need to introduce more code into [task_blueprint](https://github.com/procore/procore/blob/7862057bfaf4faad88f1c9345f27e64b989ee788/app/blueprints/schedule/task_blueprint.rb#L46) in the monolith. 
- Developers in the monolith may start using this option which will could result in blueprinter continuing to user eager loading even if we turn the flag off. This would be a problem if a bug arises and we want to roll back the eager loading changes in blueprinter. We will not be able to do so without a redeploy

lib/blueprinter/helpers/base_helpers.rb
davidyeeprocore1: Good call. I'll add a check for this

lib/blueprinter/helpers/base_helpers.rb
philipqnguyen: Blueprinter is a public gem and changes to Blueprinter should be ready for use by the greater Ruby community. 
If the goal in this PR is just to test this feature, I suggest 2 things:
a) Either have the monolith's Gemfile reference this branch
b) Or directly override the monolith's Blueprinter with these changes (basically a monkey patch).

When you have completed the test in the monolith, then we can introduce the change into Blueprinter itself.

lib/blueprinter/helpers/association_helpers.rb
```diff
@@ -0,0 +1,69 @@
...
+    # +association_name_as_symbol+:: the association to the active record relation
+    # Example would be that an Account has_many :users. This would result in a hash
+    # with {:users => :users, :user_ids => :users}
+    def collection_associated_fields_hash(object_relation, association_name_as_symbol)
+      collection_associated_hash = Hash.new
+      if object_relation[0].respond_to?(association_name_as_symbol)
+        collection_associated_hash[association_name_as_symbol] = association_name_as_symbol
+      end
+      # do our best to determine the autogenerated method name. Does not account for custom names
+      collection_ids = association_name_as_symbol.to_s[0..-2] + '_ids'
```
philipqnguyen, line 61: This took me some time to understand why you were doing `[0..-2]`.  You can instead use `association_name_as_symbol.to_s.singularize`, which is part of ActiveSupport in Rails, and this section of the code should only be executing inside Rails.

lib/blueprinter/helpers/association_helpers.rb
```diff
@@ -0,0 +1,69 @@
...
+    # +object_relation+:: the active record association whose associations we are
+    # trying to derive.
+    # +association_name_as_symbol+:: the association to the active record relation
+    # Example would be that an Account has_many :users. This would result in a hash
+    # with {:users => :users, :user_ids => :users}
+    def collection_associated_fields_hash(object_relation, association_name_as_symbol)
+      collection_associated_hash = Hash.new
+      if object_relation[0].respond_to?(association_name_as_symbol)
+        collection_associated_hash[association_name_as_symbol] = association_name_as_symbol
+      end
```
philipqnguyen, line 59: You might be able to DRY up line 56 - 59 because these are identical to line 39 - 43 in the `singular_associated_fields_hash` method above

lib/blueprinter/helpers/base_helpers.rb
```diff
@@ -31,6 +49,28 @@ def prepare_data(object, view_name, local_options)
...
 
+      def object_relation_to_hash(local_options, object, view_name)
+        field_to_association_hash = get_field_to_association_hash(object)
+        list_of_associations_to_preload = []
+        view_collection.fields_for(view_name).each do |field|
+          if preload_field?(field, field_to_association_hash, list_of_associations_to_preload, object)
+            list_of_associations_to_preload << field_to_association_hash[field.name]
+          end
+        end
+        object_relation_with_associations_loaded = object.preload(list_of_associations_to_preload)
```
philipqnguyen, line 60: Does this preload nested associations?
For example, `User.preload(posts: :comments)`?

davidyeeprocore1: No, it does not

lib/blueprinter/helpers/base_helpers.rb
davidyeeprocore1: Since we want to scope the changes to only affect the monolith, I like the idea of pointing the monolith's Gemfile to this branch. I would also like to be able to toggle this feature with a feature flag, as opposed to a config that would allow developers to use the feature in other places as we test. The idea being that we would only scope the change to a narrow part of the code (extended view of tasks controller).

I would propose:
1. Pointing the monolith to this local branch via the Gemfile as you suggested.
2. Using the feature flag as is currently implemented on this branch

This approach would decrease the likelihood of needing a redeploy in case something goes wrong because we would be able to toggle the feature flag off.

Please let me know what you think. Thank you

lessthanjacob:
Thanks for opening this PR @davidyeeprocore1, and apologies for the extended delay here. 

Looking through the proposed changes, while there's certainly benefit here, my fear is that it over-extends the responsibilities of Blueprinter, which is primarily focused on the serialization of data and less about _how_ the data is aggregated beforehand. Moving forward, ensuring that Blueprinter isn't _directly_ tied to any persistence layer will help improve maintainability and enable more use cases for applications that do not use `ActiveRecord`.

From my perspective, to ensure clear, enforceable boundaries in application code, any required preloading should be handled prior to invoking `Blueprinter`.

njbbaer:
@davidyeeprocore1 I'm going to close this for now based on @lessthanjacob's comment, but I'd be happy to reopen if there's discussion around whether preloading should be the responsibility of Blueprinter.

ritikesh:
@njbbaer are you considering taking this proposal forward? I agree with @lessthanjacob's take that this definitely extends the use-case for a serializer. Also, when you consider supporting things like Arrays, hashes and other objects like Mongo docs, this makes little sense. ActiveRecord's internal API can change with versions and we have internally burnt are fingers trying to conditional eager_load outside of core AR. It gets really tricky to manage this across rails versions etc. So I would definitely recommend to let end-users optimise for stuff like eager-loading instead of the serialiser trying to do it magically.

njbbaer:
@ritikesh I agree with you and @lessthanjacob, this greatly expands Blueprinter's scope probably more than we want. I reached out to @davidyeeprocore1 directly to see if he wants to make a case for it, but if not I'll close this PR shortly.

jhollinger:
I can understand why this isn't being merged, but Blueprinter should at least expose the introspection capabilities necessary for other tools do this. Otherwise it's just an "N+1 query generator". Would you all be open to accepting a PR like that?

ritikesh:
@jhollinger - would something like [Bullet](https://github.com/flyerhzm/bullet) be of help?

jhollinger:
@ritikesh No, never found Bullet very useful. Dev/staging data is never complete enough to find all N+1's outside of production. Blueprint already has the info necessary to do eager loading (with a little help from AR reflections maybe). If Blueprint just exposed its definitions for other tools to hook into, it would be so much more powerful.

PR #297, opened 2023-01-17
Run tests against ruby 2.2 - 3.2

toddnestor:
Blueprinter currently only runs tests against ruby versions 2.2-2.5, this adds in 2.6-3.2 to ensure Blueprinter is compatible with the latest ruby versions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,6 +46,46 @@ jobs:
     steps: *steps_spec
     docker:
       - image: ruby:2.5
+  ruby-2_6:
+    parameters:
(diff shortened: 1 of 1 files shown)


toddnestor:
I'm closing this PR as it is clear the tests fail for ruby versions 3 and up.  This is likely just an issue with the gems used during tests and not the blueprinter code itself.

The [blueprinter-rb fork](https://github.com/blueprinter-ruby/blueprinter) is compatible with the latest versions, if this gem is still being maintained it would be good to get some of those changes merged in.

PR #299, opened 2023-04-07
Devprod 98/ecosystem division

dannyporrello:
Fixing the Backstage catalog to include the System

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -21,7 +21,7 @@ metadata:
       title: Team Slack
       icon: chat
 spec:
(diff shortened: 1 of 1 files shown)


catalog-info.yaml
```diff
@@ -21,7 +21,7 @@ metadata:
       title: Team Slack
       icon: chat
 spec:
+  system: App Management | Developer Metrics | Marketplace | Sidepanel | Support Documentation
```
dannyporrello, line 24: @claudioprocore what System does blueprinter fall under? I added all of the team-dmx systems here.

claudioprocore (approved):

PR #302, opened 2023-05-02
Check if an array_like object is an Enumerable

triciatzy:
## Context
When passing an array like `ActiveModel::Errors`, I'm unable to render blueprint because it does not pass `array_like?` condition. Both `[]` and `ActiveModel::Errors` are `Enumerable`, I'm proposing to check if array is `Enumerable` instead. 

Let me know what you guys think? Thank you.

### Before
```ruby
ModelBlueprint.render(model.errors.full_messages, root: :errors)
```

### After
```ruby
ModelBlueprint.render(model.errors, root: :errors)
```

### DIFF
--- a/lib/blueprinter/helpers/type_helpers.rb
+++ b/lib/blueprinter/helpers/type_helpers.rb
@@ -7,7 +7,7 @@ module Blueprinter
     end
 
     def array_like?(object)
-      object.is_a?(Array) || active_record_relation?(object)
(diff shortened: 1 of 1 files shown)


ritikesh:
@triciatzy - we recently fixed this in the fork - https://github.com/blueprinter-ruby/blueprinter/pull/6. Pls let me know if it helps.

triciatzy:
Thanks @ritikesh

PR #308, opened 2023-07-18
Feature proposal: Honor all conditional checks config - path for sparse fields

alejandroereyes:
#### Background
At my company we make liberal use of views. But there are times that I find them constricting when trying to optimize payload sizes. I would like to have more granular control over what is returned. I propose adding a configuration to honor all conditional checks. This would allow for a lightweight implementation for sparse fields support over our API.

Example:
```ruby
  config.roll_up_conditions = true
  config.if = ->(field_name, object, local_opts) {
    if local_opts[:sparse_fields]
      local_opts[:sparse_fields].any? { |sfield| sfield.to_s == field_name.to_s }
    else
      true
    end
  }
end

class SlothBlueprint < Blueprinter::Base
  identifier :uuid

  field :name
  field :favorite_food
  field :hair_color
  field :sunglasses, name: :has_sunglasses
end

class Sloth
  attr_accessor :name, :favorite_food, :hair_color, :sunglasses
  def initialize(name, food, color, sg)
    @name = name
    @favorite_food = food
    @hair_color = color
    @sunglasses = sg
  end

  def uuid
    Random.new_seed.to_s
  end
end

my_sloth = Sloth.new('Mirabel', 'bananas', 'brown', true)

SlothBlueprint.render_as_hash(my_sloth, sparse_fields: [:has_sunglasses, :favorite_food])
# => {:favorite_food => 'bananas', :has_sunglasses => true }

SlothBlueprint.render_as_hash(my_sloth)
# => {:uuid => '21342134...', :favorite_food => 'bananas', :hair_color => 'brown', :has_sunglasses => true, :name => 'Mirabel'}
```

Checklist:

* [X] I have updated the necessary documentation
* [ ] I have updated the changelog, if necessary (CHANGELOG.md)
* [X] I have signed off all my commits as required by [DCO](../CONTRIBUTING.md#legal)
* [X] My build is green

### DIFF
--- a/README.md
+++ b/README.md
@@ -670,7 +670,7 @@ Output:
 
 
 Both the `field` and the global Blueprinter Configuration supports `:if` and `:unless` options that can be used to serialize fields conditionally.
-By default, field-level conditions override global conditions. By enabling `roll_up_conditions`, BOTH field-level and global conditions must be met in order for the field to be serialized.
+By default, field-level conditions override global conditions. By enabling `enforce_all_conditions`, BOTH field-level and global conditions must be met in order for the field to be serialized.
 
 ### Global Config Setting - if and unless
 
@@ -678,7 +678,7 @@ By default, field-level conditions override global conditions. By enabling `roll
 Blueprinter.configure do |config|
   config.if = ->(field_name, obj, _options) { !obj[field_name].nil? }
   config.unless = ->(field_name, obj, _options) { obj[field_name].nil? }
-  config.roll_up_conditions = true
+  config.enforce_all_conditions = true
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -670,7 +670,7 @@ Output:
 
 
 Both the `field` and the global Blueprinter Configuration supports `:if` and `:unless` options that can be used to serialize fields conditionally.
-By default field-level conditions override global conditions. By enabling `roll_up_conditions`, BOTH field-level and global conditions must be met in order for field to be serialized.
+By default, field-level conditions override global conditions. By enabling `roll_up_conditions`, BOTH field-level and global conditions must be met in order for the field to be serialized.
 
 ### Global Config Setting - if and unless
 
--- a/README.md
+++ b/README.md
@@ -670,6 +670,7 @@ Output:
 
 
 Both the `field` and the global Blueprinter Configuration supports `:if` and `:unless` options that can be used to serialize fields conditionally.
+By default field-level conditions override global conditions. By enabling `roll_up_conditions`, BOTH field-level and global conditions must be met in order for field to be serialized.
 
 ### Global Config Setting - if and unless
 
@@ -677,6 +678,7 @@ Both the `field` and the global Blueprinter Configuration supports `:if` and `:u
 Blueprinter.configure do |config|
   config.if = ->(field_name, obj, _options) { !obj[field_name].nil? }
   config.unless = ->(field_name, obj, _options) { obj[field_name].nil? }
+  config.roll_up_conditions = true
 end
 ```
 
@@ -690,8 +692,7 @@ class UserBlueprint < Blueprinter::Base
 end
 ```
 
-_NOTE:_ The field-level setting overrides the global config setting (for the field) if both are set.
-
+---
 </details>
 
 <details>
--- a/lib/blueprinter/field.rb
+++ b/lib/blueprinter/field.rb
@@ -18,7 +18,7 @@ module Blueprinter
     end
 
     def skip?(field_name, object, local_options)
-      if Blueprinter.configuration.roll_up_conditions
+      if Blueprinter.configuration.enforce_all_conditions
         any_global_or_field_condition_failed?(field_name, object, local_options)
       else
         return true if if_callable && !if_callable.call(field_name, object, local_options)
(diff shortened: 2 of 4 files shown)


lib/blueprinter/field.rb
```diff
@@ -25,11 +30,26 @@ def if_callable
     @if_callable = callable_from(:if)
   end
 
+  def if_callables
+    [
+      extract_callable_from(Blueprinter.configuration.if),
+      extract_callable_from(options[:if])
+
```
njbbaer, line 37:
```suggestion
```

README.md
```diff
@@ -688,13 +688,15 @@ Output:
 ---
 
 Both the `field` and the global Blueprinter Configuration supports `:if` and `:unless` options that can be used to serialize fields conditionally.
+By default field-level conditions override global conditions. By enabling `roll_up_conditions`, BOTH field-level and global conditions must be met in order for field to be serialized.
```
njbbaer, line 691:
```suggestion
By default, field-level conditions override global conditions. By enabling `roll_up_conditions`, BOTH field-level and global conditions must be met in order for the field to be serialized.
```

njbbaer (commented):
Looks good from my perspective. I agree we should be able to combine the conditionals, and `roll_up_conditions` makes sense.

njbbaer (approved):

README.md
```diff
@@ -688,13 +688,15 @@ Output:
...
 Both the `field` and the global Blueprinter Configuration supports `:if` and `:unless` options that can be used to serialize fields conditionally.
+By default, field-level conditions override global conditions. By enabling `roll_up_conditions`, BOTH field-level and global conditions must be met in order for the field to be serialized.
 
 ### Global Config Setting - if and unless
 
 ```ruby
   config.if = ->(field_name, obj, _options) { !obj[field_name].nil? }
   config.unless = ->(field_name, obj, _options) { obj[field_name].nil? }
+  config.roll_up_conditions = true
```
njbbaer, line 699: @lessthanjacob suggested `roll_up_conditions` is a bit unclear. How about `enforce_all_conditions`?

lessthanjacob: Thanks for the ping! You caught me in the middle of writing a quick comment 😆

The desired behavior makes sense to me, but the name of the configuration setting gives me slight pause. Ideally it's something that you can set and forget, but I'd challenge us to think of a name that describes the functionality a bit more explicitly! 

I could be on board with `enforce_all_conditions`. We could also codify the default behavior in the configuration, and expect that we set it to `false` to get the behavior detailed in this PR (e.g. `field_level_condition_overrides_global`).

lessthanjacob:
Looking back at this again, I think it could be worth implementing _explicit_ Sparse Field functionality here as a follow up!

alejandroereyes:
@lessthanjacob , @njbbaer Goood call on the name change to `enforce_all_conditions`, it's much clearer. I went ahead and changed it. There were some merge conflicts that I had to resolve as well.

spec/integrations/shared/base_render_examples.rb
```diff
@@ -333,6 +333,43 @@ def self.unless_method(_field_name, _object, _options)
                 include_examples 'does not serialize the conditional field'
               end
             end
+
+            context 'enforce_all_conditions is enabled' do
+              let(:local_options) { {x: 1, y: 2, v1: value, v2: other_value} }
+
+              before do
+                Blueprinter.configuration.enforce_all_conditions = true
```
kkohrt, line 341: Setting and resetting global configurations for testing is not an ideal practice.
Can this be done instead with 
```suggestion
                allow(Blueprinter).to receive(:configuration.enforce_all_conditions).and_return(true)
```
and then, as a bonus, you don't need the `after` block

spec/integrations/shared/base_render_examples.rb
```diff
@@ -333,6 +333,43 @@ def self.unless_method(_field_name, _object, _options)
...
+                Blueprinter.configuration.if = ->(field_name, object, local_opts) {
```
kkohrt, line 342: Ideally this could also be stubbed with an `allow` condition

spec/integrations/shared/base_render_examples.rb
```diff
@@ -333,6 +333,43 @@ def self.unless_method(_field_name, _object, _options)
...
+                  if local_opts[:sparse_fields]
+                    local_opts[:sparse_fields].include?(field_name.to_s)
+                  else
+                    value
+                  end
+                }
+                Blueprinter.configuration.unless = ->(_a,_b,_c) { other_value }
```
kkohrt, line 349: ditto

kkohrt (commented):
@alejandroereyes -- thanks for adding this feature!
Throwing in some ideas here.

1. I could go with `enforce_all_conditions`; or I might suggest something using the word `merge`; e.g. `merge_field_options_with_global_options`
2. Ideally there would be more complete test coverage of the different branches:
   * field unless == true but global if == true: merge_options? ? passing : skipping
   * field unless == false, but global unless == true: merge_options? ? skipping : passing
   * field if == true, but global unless == false:  merge_options? ? passing : passing
   * invalid `if` or `unless` values for either global or field options should raise specific errors
3. I do like the break up of methods into more Clean Code; but I might DRY things up some more. Suggestions in-line, below

Issue #341, opened 2023-10-03
Discussion of a stable reflection API


jhollinger:
As a follow up to the discussion in #294, things like auto-preloading of ActiveRecord associations could be done through Blueprinter extensions _if_ there was some kind of reflection API available (think ActiveRecord's `reflections` or `columns_hash`).

I've been experimenting with such an extension and have a good handle on what kind of info it needs from Blueprinter (essentially "give me all the associations for view X recursively"). But I don't want to design something overly-specific, so I'm hoping to gather some feedback. As a point of reference, here's some POC work:

This is non-invasive to Blueprinter, aside from requesting all the associations from the view. The duplication of the Blueprinter class and view isn't ideal.

```ruby
q = Widget.
  where(...).
  order(...).
  preload_blueprint(WidgetBlueprint, :extended)

render json: WidgetBlueprint.render(q, view: :extended)
```

Here's a centralized version without the duplication. `===` is used to match objects passed to `render`, and the result of the block is passed to the render pipeline. Vaguely like a centralized transformer.

```ruby
# config/initializers/blueprinter.rb

Blueprinter.configure do |config|
  config.generator = Oj
  config.on_render ActiveRecord::Relation do |scope, blueprint, view|
    scope.preload_blueprint(blueprint, view)
  end
end
```

Very curious to hear back, either about this kind of extension or extensions in general. Thoughts I've been having are:

- Building a formal reflection API vs switching existing attrs/methods from private to public?
- Are there other obvious kinds of extensions people would want to build? (Preloading for other ORMs, doc generators, etc?)

sandstrom:
Sounds like a good idea!

Apart from preloading, doc generation (which you've also mentioned) is a good candidate. Another one would be generating fixtures dynamically (for use in client-side tests).

njbbaer:
@jhollinger Can this issue be closed with the introduction of https://github.com/procore-oss/blueprinter/pull/357 and https://github.com/procore-oss/blueprinter/pull/358?

jhollinger:
A reflection API was added in 1.0; closing.

PR #357, opened 2023-11-09
Add reflection on views, fields, and associations

jhollinger:
Initial attempt at a reflection API per #341. I've built a POC ActiveRecord automagic preloader against this and #358.

Blueprinter's native views and associations have been wrapped in simpler structs or classes, as the native ones made it pretty hard to get at stuff. Could also protect against internal changes in the future.

I exposed the internal `method` and `name` as `name` and `display_name` respectively, since Rubocop (and best practices) don't allow methods named "method".

```ruby
# Returns Hash of views keyed by name
views = WidgetBlueprint.reflections

# Returns Hash of non-association fields keyed by method name
fields = views[:default].fields
fields[:name].name
=> :name
fields[:name].display_name
=> :name

# Returns Hash of associations keyed by method name
assoc = views[:default].associations
assoc[:category].name
=> :category
assoc[:category].display_name
=> :category
assoc[:category].blueprint
=> CategoryBlueprint
assoc[:category].view
=> :default
```

TODO
- Update documentation if approach is approved
- Is there any missing info we feel is critical to expose for initial release?

### DIFF
--- a/lib/blueprinter/reflection.rb
+++ b/lib/blueprinter/reflection.rb
@@ -74,7 +74,10 @@ module Blueprinter
 
       # Returns all fields or associations from included views
       def included(type)
-        @view.included_view_names.reduce({}) do |acc, view_name|
+        view_names = @view.included_view_names
+        view_names.unshift :default unless name == :default
+
+        view_names.reduce({}) do |acc, view_name|
           view = @blueprint.reflections.fetch(view_name)
           acc.merge view.send(type)
         end
--- a/lib/blueprinter/reflection.rb
+++ b/lib/blueprinter/reflection.rb
@@ -24,7 +24,7 @@ module Blueprinter
     #
     def reflections
       @reflections ||= view_collection.views.transform_values do |view|
-        View.new(self, view)
+        View.new(view.name, view_collection)
       end
     end
 
@@ -32,14 +32,11 @@ module Blueprinter
     # Represents a view within a Blueprint.
     #
     class View
-      def initialize(blueprint, view)
-        @blueprint = blueprint
-        @view = view
-      end
+      attr_reader :name
 
-      # @return [String] The view's name
-      def name
-        @view.name
+      def initialize(name, view_collection)
+        @name = name
+        @view_collection = view_collection
       end
 
       #
@@ -48,7 +45,7 @@ module Blueprinter
       # @return [Hash<Symbol, Blueprinter::Reflection::Field>]
       #
       def fields
-        @fields ||= @view.fields.each_with_object(included(:fields)) do |(_name, field), obj|
+        @fields ||= @view_collection.fields_for(name).each_with_object({}) do |field, obj|
           next if field.options[:association]
 
           obj[field.method] = Field.new(field.method, field.name, field.options)
@@ -61,25 +58,12 @@ module Blueprinter
       # @return [Hash<Symbol, Blueprinter::Reflection::Association>]
       #
       def associations
-        @associations ||= @view.fields.each_with_object(included(:associations)) do |(_name, field), obj|
+        @associations ||= @view_collection.fields_for(name).each_with_object({}) do |field, obj|
           next unless field.options[:association]
 
           blueprint = field.options.fetch(:blueprint)
           view = field.options[:view] || :default
           obj[field.method] = Association.new(field.method, field.name, blueprint, view, field.options)
-        end
-      end
-
-      private
-
-      # Returns all fields or associations from included views
-      def included(type)
-        view_names = @view.included_view_names
-        view_names.unshift :default unless name == :default
-
-        view_names.reduce({}) do |acc, view_name|
-          view = @blueprint.reflections.fetch(view_name)
-          acc.merge view.send(type)
         end
       end
     end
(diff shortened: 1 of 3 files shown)


ritikesh:
can you pls share your POC here as well for better understanding on how both of these could be stitched together?
> Initial attempt at a reflection API per https://github.com/procore-oss/blueprinter/issues/341. I've built a POC ActiveRecord automagic preloader against this and https://github.com/procore-oss/blueprinter/pull/358.

jhollinger:
@ritikesh I hesitate to share the code as it's in a private, company repo atm. I do plan to open source it into procore-oss once all this is merged and I get company approval.

The skeleton is in #358's description, but the magic is mostly in `figure_out_preloads`. In short it builds up a nested Hash for `preload`/`includes` by finding matching associations between the Blueprinter and ActiveRecord classes, recursively. This PR allows the extension to inspect the Blueprinter associations, while #358 allows the extension to tack the preloads onto the `ActiveRecord::Relation` that was passed to `render`.

ritikesh:
@jhollinger would you be able mimic that with a raw new rails demo app with a few models like posts/users/comments etc? Would just help visualise better is all :)

jhollinger:
@ritikesh Here's a gist with the relevant bits https://gist.github.com/jhollinger/9a9056b87c3bef1cb13c1032d3f3bf89

ritikesh:
thanks @jhollinger, this helps. one final question - the `preload_blueprint` is a custom patch/extension and not something that blueprinter will provide OOTB?

jhollinger:
@ritikesh Correct, that's an ActiveRecord patch the gem with the extension will make.

ritikesh:
LGTM then.

lib/blueprinter/reflection.rb
```diff
@@ -0,0 +1,84 @@
...
+      def fields
+        @fields ||= @view.fields.each_with_object(included(:fields)) do |(_name, field), obj|
+          unless field.options[:association] # rubocop:todo Style/IfUnlessModifier
+            obj[field.method] = Field.new(field.method, field.name, field.options)
+          end
+        end
+      end
+
+      #
+      # Returns a Hash of associations in this view (recursive) keyed by method name.
```
lessthanjacob, line 59: Should we consider this recursive? When I read this initially, I had thought that we might be returning the full "tree" of associations (i.e. associations contained within top-level associations). 

Or are we just reiterating here due to the implementation of `#included`?

lib/blueprinter/reflection.rb
```diff
@@ -0,0 +1,84 @@
...
+      end
+
+      private
+
+      # Recursively returns all fields or associations from included views
+      def included(type)
+        @view.included_view_names.reduce({}) do |acc, view_name|
+          view = @blueprint.views.fetch(view_name)
+          acc.merge view.send(type)
+        end
```
lessthanjacob, line 80: This makes me wonder if we could handle this higher level in `Reflection`, and avoid calling back up it from this class. In other words, can we pass in dependent `Reflection::View` instances on initialization. 

Of course, this introduces an order dependency, which I don't believe we currently enforce.

lessthanjacob (commented):
Great work on this! Left some comments, but overall the direction LGTM!

lib/blueprinter/reflection.rb
```diff
@@ -0,0 +1,84 @@
...
+      Association = Struct.new(:name, :display_name, :blueprint, :view, :options)
+
+      def initialize(blueprint, view)
+        @blueprint = blueprint
+        @view = view
+      end
+
+      # @return [String] The view's name
+      def name
+        @view.name
```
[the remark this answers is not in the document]
lessthanjacob: I was thinking that we'd make them private. It's more of a stylistic suggestion than anything, so feel free to disregard!

lib/blueprinter/reflection.rb
jhollinger: Yeah, I think I left that comment from an original implementation that was actually recursive. Will change.

lib/blueprinter/reflection.rb
jhollinger: Is the concern that we're calling methods on `Blueprinter::View` here? I guess we could pass `view.name, view.fields, view.included_view_names` to the constructor.

lessthanjacob: It's more the circular dependency we're introducing here (e.g. `Blueprinter#reflections` aggregates `Reflection::View` objects, which can then call back to `Blueprinter#reflections`). Happy to consider this a potential follow up, though!

jhollinger: @lessthanjacob I was able to re-use `ViewCollection#fields_for` in `Reflection::View`. It's quite a bit cleaner and should be easier to maintain as features are changed/added.

lessthanjacob (approved):

ritikesh (approved):

ritikesh (approved):
these seem like big changes, can you pls update version and add a changelog entry as well?

lessthanjacob:
@ritikesh We've previously been releasing independently of any particular PR (at least for the past few versions). We can discuss further in Discord if we'd like to change this, though!

jhollinger:
@peterkarman1 Sorry, I was cleaning out some old gists. The docs for reflection & extensions are in #379. And watch this space: there may be some extensions in the works...

jhollinger:
@peterkarman1 Here's the extension that came from that gist https://github.com/procore-oss/blueprinter-activerecord

PR #358, opened 2023-11-09
Add extensions with initial support for pre_render

jhollinger:
Initial attempt at a formal extension API as part of #341. I've built a POC ActiveRecord automagic preloader against this and #357.

In this proposal, an "extension" is a class inheriting from `Blueprinter::Extension` and passed to `Blueprinter.configure`. This PR defines only one extension method: `pre_render`. It intercepts whatever is passed to `.render` and returns (potentially) a modified result.

In the simplified version of my POC below, the extension figures out which associations to preload from the Blueprint view (using reflection), and returns the `ActiveRecord::Relation` with `preload` called.

```ruby
  config.extensions << MyPreloadExtension.new
end

q = Widget.
  where(...).
  preload_blueprint

# All associations in WidgetBlueprint:default get preloaded
results = WidgetBlueprint.render(q)

class MyPreloadExtension < Blueprinter::Extension
  def pre_render(object, blueprint, view, options)
    return object unless object.is_a?(ActiveRecord::Relation) && object.preload_blueprint?

    preloads = figure_out_preloads(blueprint, view, object.model)
    object.preload(preloads)
  end
end
```

### DIFF
--- a/lib/blueprinter/extensions.rb
+++ b/lib/blueprinter/extensions.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module Blueprinter
+  #
+  # Stores and runs Blueprinter extensions. An extension is any object that implements one or more of the
+  # extension methods:
+  #
+  # The Render Extension intercepts an object before rendering begins. The return value from this
+  # method is what is ultimately rendered.
+  #
+  #   def pre_render(object, blueprint, view, options)
+  #     # returns original, modified, or new object
+  #   end
+  #
+  class Extensions
+    def initialize(extensions = [])
+      @pre_renderers = []
+      extensions.each { |ext| self << ext }
+    end
+
+    def to_a
+      @pre_renderers.uniq
+    end
+
+    # Appends an extension
+    def <<(ext)
+      @pre_renderers << ext if ext.respond_to? :pre_render
+      self
+    end
+
+    # Runs the object through all Render Extensions and returns the final result
+    def pre_render(object, blueprint, view, options = {})
+      @pre_renderers.reduce(object) do |acc, ext|
+        ext.pre_render(acc, blueprint, view, options)
+      end
+    end
+  end
+end
--- a/lib/blueprinter/extensions.rb
+++ b/lib/blueprinter/extensions.rb
@@ -14,23 +14,22 @@ module Blueprinter
   #
   class Extensions
     def initialize(extensions = [])
-      @pre_renderers = []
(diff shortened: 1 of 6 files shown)


lib/blueprinter/extensions.rb
```diff
@@ -0,0 +1,38 @@
...
+      extensions.each { |ext| self << ext }
+    end
+
+    def to_a
+      @pre_renderers.uniq
+    end
+
+    # Appends an extension
+    def <<(ext)
+      @pre_renderers << ext if ext.respond_to? :pre_render
```
ritikesh, line 27: early thoughts, should we define a base class for all Extensions too? like a `Blueprinter::Extension`? This would help us enforce structure later on.

spec/units/extensions_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    before :each do
+      Blueprinter.configure do |config|
+        config.extensions = all_extensions
+      end
+    end
+
+    after :each do
+      Blueprinter.configure do |config|
+      end
+    end
```
ritikesh, line 74: what's this for? seems like a no-op

jhollinger: It resets the config to defaults (i.e. no extensions).

ritikesh: this is a hack then, aren't transactional_fixtures enabled? if they were, we wouldn't need this.

jhollinger: Transactional fixtures? I don't believe that would affect in-memory BP configuration.

ritikesh: yeah, my bad, I got mixed up with something else. 🤦 

Either ways, the snippet below from the `Configuration` class memoizes the `configuration` in memory through a class instance variable. So how exactly is `Blueprinter.configure {}` resetting the configs to defaults? It should simply return the memoized(in-memory) `configuration` right?
```ruby
  def self.configuration
    @configuration ||= Configuration.new
  end

  def self.configure
    yield configuration if block_given?
  end
```

You might want to have a look at [`SpecHelpers#reset_blueprinter_config`](https://github.com/search?q=repo%3Aprocore-oss/blueprinter%20reset_blueprinter_config!&type=code) instead - unless it's already being invoked.

jhollinger: Oh, I see what you mean now. Fixing.

ritikesh (approved):

lessthanjacob (approved):

PR #363, opened 2023-11-23
Allowing to use included views transformer from the current view

bhooshiek-narendiran:
:
Fix for [225](https://github.com/procore-oss/blueprinter/issues/225)

PR #372, opened 2023-12-26
Allow transformers to be included across views

njbbaer:
Allows transformers to be included by other views.

Branched off of https://github.com/procore-oss/blueprinter/pull/363 by @bhooshiek-narendiran and resolves https://github.com/procore-oss/blueprinter/issues/225.

- Refactored logic for gathering included transformers.
- Added a View Collection tests suite.
- Fixed handling of global default transformers and transformers of the default view.

Checklist:

* [x] My build is green


spec/units/view_collection_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+  end
+
+  describe '#[]' do
+    it 'should return the view if it exists' do
+      expect(view_collection[:default]).to eq(view_collection.views[:default])
+    end
+
+    it 'should create the view if it does not exist' do
+      expect(view_collection[:view]).to eq(view_collection.views[:view])
+    end
```
njbbaer, line 26: Consider modifying

spec/units/view_collection_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    let(:parent_view_collection) { Blueprinter::ViewCollection.new }
+
+    before do
+      parent_view_collection[:view] << new_field
+    end
+
+    it 'should inherit views from the parent' do
+      view_collection.inherit(parent_view_collection)
+      expect(view_collection[:view].fields).to eq(parent_view_collection[:view].fields)
+    end
```
njbbaer, line 49: Rename view to `parent_view`

spec/units/view_collection_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+      it 'should return the transformers for the view' do
+        expect(view_collection.transformers(:view)).to eq([transformer])
+      end
+
+      it 'should not return any transformers for another view' do
+        view_collection[:foo]
+        expect(view_collection.transformers(:foo)).to eq([])
+      end
+
+      context 'include view transformer' do
```
njbbaer, line 90: Ensure uniqueness of transformers

lib/blueprinter/view_collection.rb
```diff
@@ -35,7 +35,8 @@ def fields_for(view_name)
     end
 
     def transformers(view_name)
-      views[view_name].transformers
+      included_transformers = gather_transformers_from_included_views(view_name)
+      included_transformers.concat(views[:default].transformers).uniq
```
njbbaer, line 39: A change from https://github.com/procore-oss/blueprinter/pull/363 is that the view transformers are now merged with the default transformers, instead of replacing them:

https://github.com/procore-oss/blueprinter/blob/e3c8ecc1ae461a6f54ac5c67d0dcd53471b5b1c3/lib/blueprinter/view_collection.rb#L39

This is the behavior I expected, since default fields on the blueprint are merged with the views' fields. Though it's debatable as "default" can be ambiguous.

lib/blueprinter/view_collection.rb
```diff
@@ -35,7 +35,9 @@ def fields_for(view_name)
     end
 
     def transformers(view_name)
-      views[view_name].transformers
+      included_transformers = gather_transformers_from_included_views(view_name)
+      all_transformers = included_transformers.concat(views[:default].view_transformers).uniq
+      all_transformers.empty? ? Blueprinter.configuration.default_transformers : all_transformers
```
njbbaer, line 40: We need to make a distinction between the configured default transformers and transformers of the default view.

According to [this section](https://github.com/procore-oss/blueprinter#global-transforms) of the README, configured default transformers should only be used if there are no other transformers. That's fine, and I've moved that logic here.

But I don't believe transformers of the default view should be treated in the same way, which I've changed from how https://github.com/procore-oss/blueprinter/pull/363 handles it. Instead, there are merged with the other view transformers. This is the behavior I would expect, since fields on default view are merged with the view's other fields.

I've added tests for both of these situations.

lib/blueprinter/view.rb
```diff
@@ -16,10 +16,6 @@ def initialize(name, fields: {}, included_view_names: [], excluded_view_names: [
       @sort_by_definition = Blueprinter.configuration.sort_fields_by.eql?(:definition)
     end
 
-    def transformers
-      view_transformers.empty? ? Blueprinter.configuration.default_transformers : view_transformers
-    end
-
```
njbbaer, line 22: This logic has been moved inside of `ViewColleciton` since we need to aggregate all of the included transformers before deciding whether or not to apply the configured default transformers. See https://github.com/procore-oss/blueprinter/pull/372#discussion_r1436806378 for details.

jmeridth (approved):

lib/blueprinter/view_collection.rb
```diff
@@ -89,5 +91,14 @@ def add_to_ordered_fields(ordered_fields, definition, fields, view_name_filter =
         ordered_fields[definition.name] = fields[definition.name]
       end
     end
+
+    def gather_transformers_from_included_views(view_name)
```
lessthanjacob, line 95: **nit**:  Should we make this `private`?

spec/spec_helper.rb
```diff
@@ -2,6 +2,10 @@
 
 $LOAD_PATH.unshift File.expand_path('../../lib', __FILE__)
 require 'blueprinter'
+require 'pry'
+require 'json'
```
lessthanjacob, line 6: Are these `require`'s necessary?

lib/blueprinter/view_collection.rb
jhollinger: This is an edge case, but in what order should the transformers be applied? I would expect ones from the `default` view first, followed by included views, and finally any from the final view. Looks like this might apply the default view ones last. Thoughts?

jhollinger (commented):
Looks good, but I have some thoughts on the ordering of transformers.

lib/blueprinter/view_collection.rb
bhooshiek-narendiran: @njbbaer Currently, the default views transformer is not applied when calling with a specific view. Do we need to change this behavior?

bhooshiek-narendiran: @jhollinger I prefer from last to first in views because that provides easy tracing of transformers for the end implementer. If we taking that route can we also follow the same for default too?

spec/spec_helper.rb
m-graf: I hope not! :)

spec/units/view_collection_spec.rb
```diff
@@ -0,0 +1,142 @@
+# frozen_string_literal: true
+
+describe 'ViewCollection' do
+  let(:view_collection) { Blueprinter::ViewCollection.new }
```
m-graf, line 4: could we use `subject` here?

spec/units/view_collection_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+
+  let!(:default_view) { view_collection[:default] }
+  let!(:view) { view_collection[:view] }
```
m-graf, line 7: Why are we using `let!` here?

m-graf (commented):
small comments

lib/blueprinter/view_collection.rb
njbbaer: Isn't it already?

spec/units/view_collection_spec.rb
njbbaer: When a view collection tries to fetch a view that doesn't exist it creates it, which is what we're doing here. We need that view to exist for our tests, but don't always call `view` directly, so it won't be initialized unless we use `let!`.

lib/blueprinter/view_collection.rb
jhollinger: @bhooshiek-narendiran I may not have defined my terms well enough. I was talking about "transformers from the default view", as opposed to "the default transformers configuration setting". All discussion of "default" below means "transformers from the default view".

Allow me to amend my comment. In the PR's current code, when transformers are present, they _appear_ to be applied in this order:

```
1) Transformers from intermediate views (top-down), 2) Transformers from the named view, 3) Transformers from :default view
```

That ordering strikes me as arbitrary, and I could see it leading to confusion in some edge cases. IMHO a much more predictable ordering would be "descending":

```
1) Transformers from :default view, 2) Transformers from intermediate views (top-down), 3) Transformers from the named view
```
@bhooshiek-narendiran suggested the reverse of that order, which I also think would be fine:

```
1) Transformers from the named view, 2) Transformers from intermediate views (bottom-up), 3) Transformers from :default view.
```

spec/spec_helper.rb
njbbaer: I added `require 'json'` here because without it many spec files fail when run individually (i.e. `bundle exec rspec spec/units/view_spec.rb`) because JSON is uninitialized. The test suite works currently when run in its entirety because of the order of the specs. When we use active record in the integration tests, which run first, it imports JSON and allows the subsequent tests to pass. Is it a bad idea to import this dependency in the spec helper?

I added `require 'pry'` simply for the convenience of debugging, but I understand if that's superfluous.

lib/blueprinter/view_collection.rb
njbbaer: Thanks @jhollinger, that's an important point. I'm inclined towards the "descending" order, which gives named views the opportunity to have the final say on their rendering, so to speak.

@bhooshiek-narendiran can you elaborate on why you prefer the reverse?

> Currently, the default views transformer is not applied when calling with a specific view. Do we need to change this behavior?

Yes, we want to change this behavior so that it's easier to apply transformers more broadly. When a field is added to the default view it is applied to all of the named views, likewise a transformer added to the default view should also be applied in addition to the named views.

bhooshiek-narendiran: @njbbaer @jhollinger Apologize I also meant the same thing.
```
class Sample	
	transformer_view_default

	view_1 do
	transformer_1
	end

	view_2 do
	transformer_2
	end

	view_3 do 
	transformer_3
	end
end

```

Execution order of transformer =>` transformer_3,transformer_2,transformer_1 `
Do we need to merge **transformer_view_default** along with views transformer or only during views transformer not present.

But Adding it to first or last both seems to be a good addition.

njbbaer: @jhollinger @bhooshiek-narendiran
I've made the PR changes described, so the execution order is now "top-down".

@bhooshiek-narendiran
Not quite. In your example (without `include` statements) each view would execute these transformers in this order:
`view_1` => `transformer_view_default`, `transformer_1`
`view_2` => `transformer_view_default`, `transformer_2`
`view_2` => `transformer_view_default`, `transformer_3`

Here's a more complete example, which uses included views:
```
class Sample
  transform TransformViewDefault
  
  view :view_1 do
    transform Transformer1
  end

  view :view_2 do
    transform Transformer2
    include :view_1
  end

  view :view_3 do
    transform Transformer3
    include :view_2
  end
end
```

`view_1` => `TransformViewDefault`, `Transformer1`
`view_2` => `TransformViewDefault`, `Transformer1`, `Transformer2`
`view_2` => `TransformViewDefault`, `Transformer1`, `Transformer2`, `Transformer3`

The idea is that the "more specific" transformers get executed later, so they can get the last word on their rendering.

m-graf (approved):

jhollinger (approved):
👍

bhooshiek-narendiran (approved):

lessthanjacob (approved):

PR #379, opened 2024-01-17
Docs for reflection and extensions

jhollinger:
See the rendered updates at https://github.com/jhollinger/blueprinter/blob/ref-ext-docs/README.md.

Checklist:

* [x] My build is green

### DIFF
--- a/README.md
+++ b/README.md
@@ -933,7 +933,7 @@ Output:
 <details>
 <summary>Reflection</summary>
 
-Blueprint classes may be reflected on to inspect their views, fields, and associations.
+Blueprint classes may be reflected on to inspect their views, fields, and associations. Extensions often make use of this ability.
 
 ```ruby
 # A Hash of views keyed by name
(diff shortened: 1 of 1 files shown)


jmeridth (dismissed):

README.md
```diff
@@ -930,6 +930,71 @@ Output:
...
+fields[:description].name
+fields[:description].display_name
+fields[:description].options
+
+# Get info about an association
+assoc[:category].name
+assoc[:category].display_name
+assoc[:category].blueprint
+assoc[:category].view
+assoc[:category].options
```
ritikesh, line 958: would it help to first demonstrate or show an example of the `WidgetBlueprint` that we're playing around with? helps with better visualisations on what's happening with all these magical methods

README.md
```diff
@@ -930,6 +930,71 @@ Output:
...
+Blueprinter.configure do |config|
+  config.extensions << MyExtension.new
+  config.extensions << OtherExtension.new
+end
+```
+
+The following class implements all extension hooks. If you have an idea for a new one, open an issue or PR.
+
+```ruby
+class ExampleExtension < Blueprinter::Extension
```
ritikesh, line 977: any real-life examples that could help sparkle thought process?

README.md
```diff
@@ -930,6 +930,71 @@ Output:
...
+  # @param object [Object] The object that was passed to "render"
+  # @param blueprint [Class] The Blueprinter class
+  # @param view [Symbol] The blueprint view
+  # @param options [Hash] Options passed to "render"
+  # @return [Object] The modified, or new, object to render
+  #
+  def pre_render(object, blueprint, view, options)
+    # modify or replace object
+    object
+  end
```
ritikesh, line 993: just thinking if we should move away from the field-level type formatters in favour of extension APIs for future use-cases and provide a set of default extensions OOTB that can be configured?
DateFormatterExtension, CurrencyFormatterExtension comes to mind easily..

jhollinger: Interesting - a "blueprinter-contrib" kind of thing? I guess we could package up any number of extensions and/or formatters.

README.md
jhollinger: https://github.com/procore-oss/blueprinter-activerecord

ritikesh: Should we add this link to the readme as well for reference?

README.md
```diff
@@ -930,6 +930,85 @@ Output:
...
+  config.extensions << OtherExtension.new
+end
+```
+
+The following class implements all extension hooks. If you have an idea for a new one, open an issue or PR.
+
+```ruby
+class ExampleExtension < Blueprinter::Extension
+  #
+  # Called eary during "render", this method receives the object to be rendered and
```
lessthanjacob, line 989: I think we missed this in the top level comment of `Blueprinter::Extension` as well. 

```suggestion
  # Called early during "render", this method receives the object to be rendered and
```

lessthanjacob: As an aside, it looks like we're essentially just copying the base class here. Would it be helpful to either:

1. Link out to the code itself, and provide a list of "hooks" that are currently available
2. Enumerate the list of "hooks" and add an example using them.

sandstrom:
Great with reflection docs! I was just trying to find them, and then happily found this PR. Thanks! 😄

ritikesh (approved):

jmeridth (approved):

Issue #281, opened 2021-09-21
Make field naming more consistent


jasonkarns:
Fields that are simply "renamed" use: `field :internal_name, name: :output_name`

But fields that are defined "inline" use `field(:output_name) { ... }`

In the former case, the argument to `field` is _not_ the final output name, but in the latter case it is. I find this inconsistency a bit jarring. I understand why it is this way, but it feels backwards to me.

What would be the appetite for an API that did something like: (`method` name adopted from graphql-ruby's parameter for the same behavior: essentially aliasing.)

```
field :out_name, method: :internal_name
```

This way the parameter to `field` is consistently the _resulting_ key, regardless if the field's key name remains unchanged; if it is renamed, or if it is defined directly in the blueprint.

franzliedke:
I came here to suggest exactly this.

When I look at the serializer, I want to be able to quickly see the keys of the resulting JSON object at a glance, without having to look at options.

I am happy to open a PR with this feature, but would like to know whether you would accept this idea first.

The API suggested by @jasonkarns would allow for a backward-compatible change: using the old `name:` keyword would still work, but be marked deprecated, and the `method:` keyword (or something else, maybe more generic to not prescribe the method of extraction).

ritikesh:
@jasonkarns this is a good suggestion - would you be willing to raise a PR for it? Let's ensure backward compatibility is maintained though with a deprecation notice for the older approach.

franzliedke:
I'm also happy to work on such a PR. 😇

jhollinger:
I agree that this proposal makes more sense than what we have today. I know @ritikesh suggested deprecating the old way, but that implies eventually removing the old way, and that could be a significant challenge for large code bases.

To avoid that, what does everyone think about supporting _both_ ways? (Which to give credit where credit's due, was @njbbaer's idea iirc).

```ruby
field :foop, name: :foop  # a field called "foo" that pulls from a "foop" method
field :bar, method: :barp # a field called "bar" that pulls from a "barp" method
```

Granted it's a bit confusing using them _right beside_ each other, but that's on whoever writes the code. Presumably an entire app, or at least an entire view or Blueprint, would choose one way and stick with it.

Also, I don't love using `method` since there's a [`method` instance method on  `Object`](https://ruby-doc.org/3.2.2/Object.html#method-i-method). Can anyone think of a better name?

njbbaer:
How about `accessor` or `source` as an alternative to `method`?

[...]

PR #389, opened 2024-01-26
Do not overwrite overridden fields and associations

elliothursh:
# Description
First off, thank y'all for adding this new reflection feature. It's super useful! I ran into an issue with fields and categories that exist in the default view but are overridden in other views. Reflections seem to get overwritten by their definitions in the default view instead of the other view. This PR changes the behavior to not overwrite overridden fields and associations with definitions in the default view.

Checklist:

- [X] I have updated the necessary documentation
- [x] I have signed off all my commits as required by [DCO](https://github.com/procore-oss/blueprinter/blob/main/CONTRIBUTING.md)
- [x] My build is green

### DIFF
--- a/lib/blueprinter/reflection.rb
+++ b/lib/blueprinter/reflection.rb
@@ -48,7 +48,7 @@ module Blueprinter
         @fields ||= @view_collection.fields_for(name).each_with_object({}) do |field, obj|
           next if field.options[:association]
 
-          obj[field.method] = Field.new(field.method, field.name, field.options)
+          obj[field.method] ||= Field.new(field.method, field.name, field.options)
         end
       end
 
@@ -63,7 +63,7 @@ module Blueprinter
 
           blueprint = field.options.fetch(:blueprint)
           view = field.options[:view] || :default
-          obj[field.method] = Association.new(field.method, field.name, blueprint, view, field.options)
+          obj[field.method] ||= Association.new(field.method, field.name, blueprint, view, field.options)
         end
       end
     end
--- a/spec/units/reflection_spec.rb
+++ b/spec/units/reflection_spec.rb
@@ -45,6 +45,11 @@ describe Blueprinter::Reflection do
       view :legacy do
         association :parts, blueprint: part_bp, name: :pieces
       end
+
+      view :aliased_names do
+        field :name, name: :aliased_name
+        association :category, blueprint: cat_bp, name: :aliased_category
+      end
     end
   }
 
@@ -56,6 +61,7 @@ describe Blueprinter::Reflection do
       :extended_plus,
       :extended_plus_plus,
       :legacy,
+      :aliased_names
     ].sort
   end
 
@@ -76,6 +82,17 @@ describe Blueprinter::Reflection do
     ].sort
   end
 
+  it 'should list overridden fields' do
+    fields = widget_blueprint.reflections.fetch(:aliased_names).fields
+    expect(fields.keys.sort).to eq [
+      :id,
+      :name,
+    ].sort
+    name_field = fields[:name]
+    expect(name_field.name).to eq :name
+    expect(name_field.display_name).to eq :aliased_name
+  end
+
   it 'should list associations' do
     associations = widget_blueprint.reflections.fetch(:default).associations
     expect(associations.keys).to eq [:category]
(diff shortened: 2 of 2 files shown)


elliothursh:
@jhollinger I also ran into another issue around duplicate field aliases, but I couldn't think of a good way to solve in the reflection code without a breaking change and have a workaround. I think it's a pretty big edge case, but wondering if you had any thoughts?

In a blueprint defined like:
```
class FooBlueprint < Blueprinter::Base
  field :name, name: :label
  field :name, name: :value
end
```
The field second field doesn't get defined in the reflection's fields:
```
FooBlueprint.reflections[:default].fields
=> {:name=>#<struct Blueprinter::Reflection::Field name=:name, display_name=:value, options={:name=>:value, :block=>nil}>}
```
Even though it's defined in the Reflection::View.
```
FooBlueprint.reflections[:default]
=>
#<Blueprinter::Reflection::View:0x000000010f673530
 @fields={:name=>#<struct Blueprinter::Reflection::Field name=:name, display_name=:value, options={:name=>:value, :block=>nil}>},
 @name=:default,
 @view_collection=
  #<Blueprinter::ViewCollection:0x000000010f674020
   @sort_by_definition=false,
   @views=
    {:identifier=>#<Blueprinter::View:0x000000010ffb1bd8 @definition_order=[], @excluded_field_names=[], @fields={}, @included_view_names=[], @name=:identifier, @sort_by_definition=false, @view_transformers=[]>,
     :default=>
      #<Blueprinter::View:0x000000010ffb1b88
       @definition_order=[],
       @excluded_field_names=[],
       @fields=
        {:value=>
          #<Blueprinter::Field:0x000000010ffb1ae8
           @blueprint=FooBlueprint,
           @extractor=...,
           @method=:name,
           @name=:value,
           @options={:name=>:value, :block=>nil}>,
         :label=>
          #<Blueprinter::Field:0x000000010ffb1a48
           @blueprint=FooBlueprint,
           @extractor=...,
           @method=:name,
           @name=:label,
           @options={:name=>:label, :block=>nil}>},
       @included_view_names=[],
       @name=:default,
       @sort_by_definition=false,
       @view_transformers=[]>}>>
```

I think in order to resolve, I'd have to change the fields/associations to be keyed off of `field.name` instead of `field.method` which would break existing usage for a lot of people.

jhollinger:
Good catch, thanks! However, I'm surprised that [ViewCollection#fields_for](https://github.com/procore-oss/blueprinter/blob/81ae047cffd265913c94f273f8df16f36206566d/lib/blueprinter/view_collection.rb#L26) is returning both the default and overridden of field. Other uses of `fields_for` aren't doing this kind of check [[1](https://github.com/procore-oss/blueprinter/blob/81ae047cffd265913c94f273f8df16f36206566d/lib/blueprinter/helpers/base_helpers.rb#L48)], [[2](https://github.com/procore-oss/blueprinter/blob/81ae047cffd265913c94f273f8df16f36206566d/lib/blueprinter/helpers/base_helpers.rb#L121)]. Is there something different about those use cases I'm not seeing, or is the bug we need to fix actually in `fields_for`?

jhollinger:
@elliothursh Your other comment reminds me of #281 (which is old, but there have been recent discussions around it). It would probably make more sense for the reflection hashes to be keyed off of `name`, not `method`. Agreed it's technically a breaking change, but it's also broken as-is, so...yeah.

elliothursh:
I *think* fields for is working correctly. 

With a Blueprint defined like this:
```
class FooBlueprint < Blueprinter::Base
  field :name

  view :bar do
    field :name, name: :label
    field :name, name: :value
  end
end
```

It outputs this:
```
FooBlueprint.render_as_hash(OpenStruct.new(name: "NAME"), view: :bar)
=> {:label=>"NAME", :name=>"NAME", :value=>"NAME"}
```

I think the the `Blueprinter::Views.fields_for` behavior is correct to also return the default fields fields in other views. I think it works because in Blueprinter::Views fields are keyed off of the field `name`. So I kind of think the main issue is just that reflections are being keyed off of `method`.


Another example:
```
class  FooBlueprint < Blueprinter::Base
  field :name

  view :bar do
    field :baz, name: :name
    field :name, name: :value
  end
end
```
I'd expect that the render output json overrides the `name` key with the value in `:baz`, which does happen.

```
FooBlueprint.render_as_hash(OpenStruct.new(name: "NAME", baz: "BAZ"), view: :bar)
=> {:name=>"BAZ", :value=>"NAME"}
```

elliothursh:
I made a new branch with the change use field.name, and added tests for aliases (first example above) and overrides (second example above). 
https://github.com/procore-oss/blueprinter/compare/main...elliothursh:blueprinter:reflection-use-name?expand=1

Not sure how/if ya'll want to move forward with a breaking change like this, but let me know!

jhollinger:
@elliothursh The reflection feature is new enough (a couple weeks) and [currently undocumented enough](https://github.com/procore-oss/blueprinter/pull/379), that I'd personally be okay with calling the "breaking change" in your [reflection-use-name](https://github.com/procore-oss/blueprinter/compare/main...elliothursh:blueprinter:reflection-use-name?expand=1) branch a "bugfix with notes." Curious how others feel.

elliothursh:
If you're okay with calling it a 'bugfix with notes', so am I! I'll close this PR and open one for reflection-use-name instead.

PR #391, opened 2024-01-31
Use field.name when building reflective views and associations

elliothursh:
# Description
This is a followup PR to #389, that fixes the problem at hand in a different (but more correct) way.

I ran into an issue with fields and categories that exist in the default view but are overridden in other views. Reflections seem to get overwritten by their definitions in the default view instead of the other view. This PR changes the behavior to key field and association reflections off of their `name` instead of their `method`, which the same way that Blueprinter uses when serializing data.

Checklist:

* [x] My build is green

### DIFF
--- a/lib/blueprinter/reflection.rb
+++ b/lib/blueprinter/reflection.rb
@@ -48,7 +48,7 @@ module Blueprinter
         @fields ||= @view_collection.fields_for(name).each_with_object({}) do |field, obj|
           next if field.options[:association]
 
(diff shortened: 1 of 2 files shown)


jhollinger (approved):
LGTM. I know this is technically a breaking change for some use cases (i.e. when using custom field names), but it's also an important bug fix.

lessthanjacob (approved):
Thanks for putting this fix together!

PR #415, opened 2024-05-25
docs: update Transform examples

SaxtonDrey:
While learning how to use Blueprinter, I encountered some confusion with the current code examples. 
The concepts are well-explained, but there appears to be a discrepancy with the method names used in the examples. 

Checklist:

* [x] My build is green

### DIFF
--- a/README.md
+++ b/README.md
@@ -782,10 +782,10 @@ end
 ```ruby
 class User
   def custom_columns
-    self.dynamic_fields #which is an array of some columns
+    # which is an array of some columns
   end
 
-  def custom_fields
+  def dynamic_fields
     custom_columns.each_with_object({}){|col,result|  result[col] = self.send(col)}
   end
 end
--- a/README.md
+++ b/README.md
@@ -773,7 +773,7 @@ Create a Transform class extending from `Blueprinter::Transformer`
 
 ```ruby
 class DynamicFieldTransformer < Blueprinter::Transformer
-  def transform(hash, object, options)
+  def transform(hash, object, _options)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -782,10 +782,10 @@ end
 ```ruby
 class User
   def custom_columns
-    self.dynamic_fields #which is an array of some columns
+    # which is an array of some columns
   end
 
-  def custom_fields
+  def dynamic_fields
```
SaxtonDrey, line 788: aligned to `hash.merge!(object.dynamic_fields)` in [here](https://github.com/procore-oss/blueprinter/pull/415/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R777)

lessthanjacob (commented):
Thanks for opening this!

Reading through these docs again, I think this example is unnecessarily confusing generally. We could remove the `custom_columns` method entirely, and rework `dynamic_fields` to make this a bit easier to follow (i.e. more explicit). Something like:

```ruby
class User
  def dynamic_fields
    case role
    when :admin
      { 
        employer: employer,
        time_in_role: determine_time_in role
      }
    when :maintainer
      { 
        label: label,
        settings: generate_settings_hash
      }
    when :read_only
      { 
        last_login_at: last_login_at
      }
    end
  end
end

SaxtonDrey:
Thanks for your review and suggestion!
I agree that the current example is too complicated, so I updated PR(I just only took in your suggestion 😄 ).
Let me know if there are any further adjustments needed.

lessthanjacob (approved):

PR #423, opened 2024-06-12
Conditional views

alexkalderimis:
Prior to this change, applying conditions to a view required annotating
all fields individually.

This change allows views to be defined as `view :name, if: condition`,
using the same `->(field_name, object, options) { ... }` callbacks as
used on fields. These callbacks are propagated to the fields, and are in
addition to (not in replacement of) any checks defined on the fields.

Tests are added to cover the logic of `unless` and `if`, and
`Field#skip?` is altered to allow a field to have both `unless` and `if`
defined.

Defining a view-level condition on an child blueprint does not affect
the parent view of the same name - this behavior may be unexpected, and
it is advised not to add conditions to inherited views, since this may
represent a security flaw.

Checklist:

* [ ] My build is green

### DIFF
--- a/lib/blueprinter/field.rb
+++ b/lib/blueprinter/field.rb
@@ -18,9 +18,34 @@ module Blueprinter
     end
 
     def skip?(field_name, object, local_options)
+      return true if unless_callable && unless_callable.call(field_name, object, local_options)
       return true if if_callable && !if_callable.call(field_name, object, local_options)
 
-      unless_callable && unless_callable.call(field_name, object, local_options)
+      false
+    end
+
+    def clone
+      opts = options.dup
+      opts[:if] = if_callable if if_callable
+      opts[:unless] = unless_callable if unless_callable
+
+      Field.new(@method, @name, @extractor, @blueprint, opts)
+    end
+
+    # Adds an outer scoped if to any potential if conditions
+    def add_if(condition)
+      return unless condition
+
+      @if_callable = nil
+      inner = callable_from(:if)
+      outer = coerce_callable(condition)
+
+      @if_callable = lambda do |name, obj, opts|
+        next false if inner && !inner.call(name, obj, opts)
+        next false unless outer.call(name, obj, opts)
+
+        true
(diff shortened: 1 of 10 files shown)


lib/blueprinter/field.rb
```diff
@@ -18,9 +18,34 @@ def extract(object, local_options)
...
 
     def skip?(field_name, object, local_options)
+      return true if unless_callable && unless_callable.call(field_name, object, local_options)
       return true if if_callable && !if_callable.call(field_name, object, local_options)
 
-      unless_callable && unless_callable.call(field_name, object, local_options)
+      false
+    end
+
+    def clone
```
alexkalderimis, line 27: This is necessary to ensure we don't mutate parent fields by adding conditions to them

lib/blueprinter/view.rb
```diff
@@ -22,20 +31,20 @@ def track_definition_order(method, viewable: true)
...
         include_view(view_name)
       end
 
-      view.excluded_field_names.each do |field_name|
+      parent.excluded_field_names.each do |field_name|
         exclude_field(field_name)
       end
 
-      view.view_transformers.each do |transformer|
+      parent.view_transformers.each do |transformer|
```
alexkalderimis, line 47: Variable name changed to make the intent clearer

lessthanjacob (commented):
Thanks for opening this PR and contributing back to the library! 

I'll let the other maintainers chime in here as well, but for the time being, I'd push towards holding off on incorporating this. We're currently looking into altering how configuration works generally, which _should_ end up making this sort of functionality a bit easier to reason with and implement. Until then, though, I'm a bit wary of the additional complexity this brings. 

Again, I'd like to get some additional opinions on this from the rest of the maintainers, though!

jhollinger:
I concur with @lessthanjacob. But I'd also like to understand the use-case better. Why would you want to make all fields in a view conditional? Couldn't that check be moved to the caller, where it chooses which view to render?

Issue #424, opened 2024-06-13
[With PR][Feature Request] Discard blank attribute


jamesst20:
PR: https://github.com/procore-oss/blueprinter/pull/425

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

I can talk about my specific use case scenario. I'm using Blueprinter in a Ruby on Rails application. I have created an `ApplicationBlueprint` that extends `Blueprinter::Base` and I make all my blueprints extends from that class. It's pretty much the same pattern we use for ActiveRecord or Controllers or ApplicationJob, etc.

For my form validations, I always include in my output an `errors` field for every record or associations.

Exemple of output

```javascript
{
  id: 1,
  name: null,
  errors: { name: ["can't be blank"] }
  cars: [
    { id: 1, name: "My super fast car", year: null, errors: [{year: ["can't be blank"}] }
  ]
}
```

This is done in my ApplicationBlueprint which all my blueprint extends of. A very important note here is that in Rails, calling `user.errors` will give me all errors of user but also includes all the ones from cars. In this payload, I only have interest in the error of the same 'level' i'm currently blueprinting. In order to do that, I simply reject all errors that contains a dot so my user error are basically this `{ "name": [...], "cars.year": [...] }.reject(...all attribute with dot)`

``` ruby
field(:errors) do |record|
    record.try(:errors)&.as_json&.reject { |attribute| attribute.to_s.include?(".") }.presence
end
```

### Describe the feature you'd like to see implemented

If there are no errors in a record or I am rendering something that doesn't even implement `errors` as a method, I would like to not render the field at all. I understand there is already a "similar" feature implemented:

```ruby
if = ->(field_name, obj, _options) { true || false }
```

however this would force me to compute twice the value of the field to know if I should render it or not like so:

```ruby
field(:errors, if: ->((_, obj, __) {  obj.try(:errors)&.as_json&.reject { |attribute| attribute.to_s.include?(".") }.present? } ) do |record|
  record.try(:errors)&.as_json&.reject { |attribute| attribute.to_s.include?(".") }.presence
end
```

**What I would like:**

```ruby
field(:errors, discard_nil: true) do |record|
  record.try(:errors)&.as_json&.reject { |attribute| attribute.to_s.include?(".") }.presence
end
```

bfrey08:
Thank you for this. I came here looking to develop the same functionality and was pleasantly surprised that you already did it.

PR #425, opened 2024-06-13
Feature: Discard nil value field

jamesst20:
Closes https://github.com/procore-oss/blueprinter/issues/424

Checklist:

* [OK] I have updated the necessary documentation
* [OK] I have signed off all my commits as required by [DCO](https://github.com/procore-oss/blueprinter/blob/main/CONTRIBUTING.md)
* [OK] My build is green

### DIFF
--- a/README.md
+++ b/README.md
@@ -695,7 +695,7 @@ _NOTE:_ The field-level setting overrides the global config setting (for the fie
 </details>
 
 <details>
-<summary Exclude Fields with nil Values</summary>
+<summary>Exclude Fields with nil Values</summary>
 
 
 By default, fields with `nil` values are included when rendering. You can override this behavior by setting `:exclude_if_nil: true` in the field definition. 
--- a/README.md
+++ b/README.md
@@ -695,6 +695,36 @@ _NOTE:_ The field-level setting overrides the global config setting (for the fie
 </details>
 
 <details>
+<summary Exclude Fields with nil Values</summary>
+
+
+By default, fields with `nil` values are included when rendering. You can override this behavior by setting `:exclude_if_nil: true` in the field definition. 
+
+Usage:
+
+```ruby
+class UserBlueprint < Blueprinter::Base
+  identifier :uuid
+
+  field :name
+  field :birthday, exclude_if_nil: true
+end
+
+user = User.new(name: 'John Doe')
+puts UserBlueprint.render(user)
+```
+
+Output:
+
+```json
+{
+  "name": "John Doe"
+}
+```
+
+</details>
+
+<details>
 <summary>Custom Formatting for Dates and Times</summary>
 
 
--- a/lib/blueprinter/helpers/base_helpers.rb
+++ b/lib/blueprinter/helpers/base_helpers.rb
@@ -48,7 +48,11 @@ module Blueprinter
         result_hash = view_collection.fields_for(view_name).each_with_object({}) do |field, hash|
           next if field.skip?(field.name, object, local_options)
 
-          hash[field.name] = field.extract(object, local_options)
+          value = field.extract(object, local_options)
+
+          next if value.nil? && field.options[:discard_nil]
+
+          hash[field.name] = value
         end
         view_collection.transformers(view_name).each do |transformer|
           transformer.transform(result_hash, object, local_options)
--- a/lib/blueprinter/helpers/base_helpers.rb
+++ b/lib/blueprinter/helpers/base_helpers.rb
@@ -50,7 +50,7 @@ module Blueprinter
 
           value = field.extract(object, local_options)
 
-          next if value.nil? && field.options[:discard_nil]
+          next if value.nil? && field.options[:exclude_if_nil]
 
           hash[field.name] = value
         end
--- a/spec/integrations/base_spec.rb
+++ b/spec/integrations/base_spec.rb
@@ -82,6 +82,30 @@ describe '::Base' do
       end
     end
 
+    context 'Given exclude_if_nil is passed' do
+      let(:obj) { OpenStruct.new(obj_hash.merge(nil_attribute: nil)) }
+
+      context 'and exclude_if_nil is true' do
+        let(:blueprint) do
+          Class.new(Blueprinter::Base) do
+            field :nil_attribute, exclude_if_nil: true
+          end
+        end
+        let(:result) { '{}' }
+        it { expect(blueprint.render(obj)).to eq(result) }
+      end
+
+      context 'and exclude_if_nil is false' do
+        let(:blueprint) do
+          Class.new(Blueprinter::Base) do
+            field :nil_attribute, exclude_if_nil: false
+          end
+        end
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -694,6 +694,35 @@ _NOTE:_ The field-level setting overrides the global config setting (for the fie
 
 </details>
 
+<details>
+<summary>Discarding nil fields</summary>
```
lessthanjacob, line 698: **suggestion**: Slight nitpick, but I'd suggest framing this as "omit" or "exclude" instead of "discard". 

```suggestion
<summary Exclude Fields with nil Values</summary>
```

README.md
```diff
@@ -694,6 +694,35 @@ _NOTE:_ The field-level setting overrides the global config setting (for the fie
...
+
+
+By default nil values are included. You can override this behavior by passing `:discard_nil` option.
```
lessthanjacob, line 701:
```suggestion
By default, fields with `nil` values are included when rendering. You can override this behavior by setting `:exclude_if_nil: true` in the field definition. 
```

README.md
```diff
@@ -694,6 +694,35 @@ _NOTE:_ The field-level setting overrides the global config setting (for the fie
...
+
+```ruby
+class UserBlueprint < Blueprinter::Base
+  identifier :uuid
+
+  field :name
+  field(:birthday, discard_nil: true) do |_user, options|
+    nil
+  end
+end
```
lessthanjacob, line 713: Let's simplify this a bit:

```suggestion
Usage:

```ruby
  identifier :uuid

  field :name
  field :birthday, exclude_if_nil: true
end

user = User.new(name: 'John Doe')
puts UserBlueprint.render(user)
```

lib/blueprinter/helpers/base_helpers.rb
```diff
@@ -48,7 +48,11 @@ def object_to_hash(object, view_name:, local_options:)
         result_hash = view_collection.fields_for(view_name).each_with_object({}) do |field, hash|
           next if field.skip?(field.name, object, local_options)
 
-          hash[field.name] = field.extract(object, local_options)
+          value = field.extract(object, local_options)
+
+          next if value.nil? && field.options[:discard_nil]
```
lessthanjacob, line 53:
```suggestion
          next if value.nil? && field.options[:exclude_if_nil]
```

spec/integrations/shared/base_render_examples.rb
```diff
@@ -50,6 +50,33 @@
     it('returns json with a renamed field') { should eq(result) }
   end
 
+  context 'Given blueprint has ::fields with :discard_nil argument' do
+    context 'Given :discard_nil argument is true' do
```
lessthanjacob, line 54:
```suggestion
  context 'Given blueprint has ::fields with :exclude_if_nil set' do
    context 'when :exclude_if_nil is true' do
```

spec/integrations/shared/base_render_examples.rb
```diff
@@ -50,6 +50,33 @@
...
+      let(:result) { '{"first_name":"Meg"}' }
+      let(:blueprint) do
+        Class.new(Blueprinter::Base) do
+          field :first_name, discard_nil: true
+          field :my_field, discard_nil: true do
+            nil
+          end
```
lessthanjacob, line 61: **suggestion**: Can we alter this test so that the value on the provided `object` is `nil`, instead of explicitly set as part of the `Blueprint` itself (primarily to mirror a real-world scenario as best as possible).

spec/integrations/shared/base_render_examples.rb
```diff
@@ -50,6 +50,33 @@
...
+        end
+      end
+      it('returns json without nil value fields') { should eq(result) }
+    end
+    context 'Given :discard_nil argument is false' do
```
lessthanjacob, line 66:
```suggestion
    context 'when :exclude_if_nil is false' do
```

lessthanjacob (commented):
Thank you for opening this and helping add some additional configurability to Blueprinter!

Most of my comments are primarily around naming, but happy to discuss further!

spec/integrations/shared/base_render_examples.rb
jamesst20: @lessthanjacob Yes we could, but that would mean moving the spec outside of base_render_exemples to a place like base_spec because blueprints are defined per test it seem

jamesst20: @lessthanjacob I moved the spec but i'm not sure it really is better because the `base_spec` is more about specific scenario like "is rails context" or "is hash", or .... `exclude_if_nil` is something that applies to everything. I made sure it would be easy to revert by making a different commit in case I was wrong

jamesst20:
> Thank you for opening this and helping add some additional configurability to Blueprinter!
> 
> Most of my comments are primarily around naming, but happy to discuss further!

Thanks for the quick review! Very much appreciated. I have applied all of your recommandations, after all I don't know much about this project conventions so I will stick to yours :)

spec/integrations/base_spec.rb
```diff
@@ -82,6 +82,30 @@
...
+      context 'and exclude_if_nil is false' do
+        let(:blueprint) do
+          Class.new(Blueprinter::Base) do
+            field :nil_attribute, exclude_if_nil: false
+          end
+        end
+        let(:result) { '{"nil_attribute":null}' }
+        it { expect(blueprint.render(obj)).to eq(result) }
+      end
+    end
```
lessthanjacob, line 107: **suggestion**: Just to make this a bit more deterministic:

```suggestion
    context 'Given exclude_if_nil is passed' do
      let(:obj) { OpenStruct.new(obj_hash.merge(category: nil, label: 'not nil')) }

      context 'and exclude_if_nil is true' do
        let(:blueprint) do
          Class.new(Blueprinter::Base) do
            field :category, exclude_if_nil: true
            field :label, exclude_if_nil: true
          end
        end
        let(:result) { '{"label":"not nil"}' }
        it { expect(blueprint.render(obj)).to eq(result) }
      end

      context 'and exclude_if_nil is false' do
        let(:blueprint) do
          Class.new(Blueprinter::Base) do
            field :category, exclude_if_nil: false,
            field :label, exclude_if_nil: true
          end
        end
        let(:result) { '{"category":null,"label":"not nil"}' }
        it { expect(blueprint.render(obj)).to eq(result) }
      end
    end
```

lessthanjacob (dismissed):

jamesst20:
Hey @lessthanjacob 

May I kindly bump this and ask what's the next step now that is has your approval?

lessthanjacob:
Hey @jamesst20! I believe I may have suggested some changes to the tests that weren't totally valid syntax-wise. Do you mind addressing those locally and pushing up a fix? Apologies for that!

jamesst20:
@lessthanjacob It's done! :) Indeed there was an extra `,` at the end of a line

lessthanjacob (approved):

PR #427, opened 2024-06-19
Update README.md

hack3rvaillant:
I may be wrong but I was confused by the exemple.
So I tried to fix it.
Hope this helps

Checklist:

* [x] My build is green


lessthanjacob:
Thanks for opening this PR @hack3rvaillant!

I agree that the current example is confusing, but it will change as the result of #415! Let me know if that provides a clear example!

PR #428, opened 2024-06-24
Remove Custom Array-like Class Configuration

lessthanjacob:
## Changelog
- Removes custom array-like configuration in favor of depending on an interface (`Enumerable` in this case). `Hash` is an exception here due to existing implementation details. 

## Discussion
This was [previously discussed](https://github.com/blueprinter-ruby/blueprinter/pull/2) as an option, but was decided against  due to potentially be _too_ open. 

Checklist:

* [X ] I have updated the necessary documentation
* [X ] I have signed off all my commits as required by [DCO](https://github.com/procore-oss/blueprinter/blob/main/CONTRIBUTING.md)
* [ ] My build is green

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,24 +91,6 @@ This will result in JSON that looks something like this:
   }
 ]
 ```
-
-
-You can also configure other classes to be treated like collections. For example, if you are using Mongoid, you can configure it to treat `Mongoid::Criteria` objects as collections:
-
-```ruby
-Blueprinter.configure do |config|
-  config.custom_array_like_classes = [Mongoid::Criteria]
-end
-```
-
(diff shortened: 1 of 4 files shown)


jhollinger:
I like the direction here. But I do think that these kinds of heuristics will always fail _someone_. So IMHO the important thing is to allow the user to recover from that. What if `render` called out to `render_one` or `render_all`? Then if `render` gets it wrong for \<insert obscure collection object\>, the user could call `render_one` or `render_all` directly. We'd also need a 
"hint option" for associations...

FWIW, `render_one`/`render_all` may not be the most descriptive names. `render_object` and `render_collection`?

lessthanjacob:
Will revisit at a later time.

Issue #404, opened 2024-03-21
Extractor configurability


sandstrom:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

Association extractor is useful to modify to e.g. add a circuit-breaker, implement caching and for a few other things. Right now it's somewhat complicated to override/modify.

### Describe the feature you'd like to see implemented

Right now, there are two types of extractors: AssociationExtractor and AutoExtractor (along with some sub-extractors).

Their roles are somewhat intertwined (AssociationExtractor calls the AutoExtractor, for example), and only one of them can be modified globally.

I'd propose a stricter separation, splitting it up into (probably) three extractors:

- FieldExtractor
- AssociationExtractor
- ValueExtractor (or DataExtractor)

The default `ValueExtractor` would encapsulate the logic currently existing in AutoExtractor, and I'd move the logic of deferring to block to field/association extractors. `ValueExtractor` could also be named something like `DataExtractor` instead.

## Example

This is a rough sketch, in reality it would probably be slightly different.

```ruby
class ValueExtractor < Extractor
  def extract(field_name, object, local_options, options = {})
    if object.is_a?(Hash)
      value = extract_hash(field_name, object, local_options, options)
    else
      value = extract_send(field_name, object, local_options, options)
    end
  end

  private

  def extract_hash(field_name, object, _local_options, _options = {})
    object[field_name]
  end

  def extract_send(field_name, object, _local_options, _options = {})
    object.public_send(field_name)
  end
end

class FieldExtractor < Extractor
  include EmptyTypes

  def initialize
    @extractor = Blueprinter.configuration.field_extractor_default.new # `ValueExtractor.new` by default
  end

  def extract(field_name, object, local_options, options = {})
    if options[:block]
      value = options[:block].call(object, local_options)
    else
      value = @value_extractor.extract(field_name, object, local_options, options)
    end

    value = @datetime_formatter.format(value, options)
    use_default_value?(value, options[:default_if]) ? default_value(options) : value
  end

  private

  def default_value(field_options)
    field_options.key?(:default) ? field_options.fetch(:default) : Blueprinter.configuration.field_default
  end
end

class AssociationExtractor < Extractor
  include EmptyTypes

  def initialize
    @extractor = Blueprinter.configuration.field_extractor_default.new # `ValueExtractor.new` by default
  end

  def extract(association_name, object, local_options, options = {})
    options_without_default = options.except(:default, :default_if)
    
[...]

PR #437, opened 2024-07-09
2.0 Base Class Proposal

jhollinger:
A proposed V2 base class.

#### Overview

- Combines concepts of "blueprint" and "view" (from #430).
  - A view is just a subclass of the Blueprint.
  - A nested view is a sublcass of its parent view.
  - The `:default` view is just an alias to `self` (i.e. the Blueprint class).
- `object` = singular association, `collection` = multiple (names up for debate)
- Fields, partials, extensions, and options are inherited from the parent class/view, but can then be customized.
- Fields and associations are just structs.
- "Partials" can be defined and included in views.

#### Notes

- There's no `MyBlueprint < Blueprinter[2.0]` syntax in this PR, but we can add it later if we want.
- More options for views and fields/associations will need to be added in later PRs.

### Defining

```ruby
# "global" config goes here
class ApplicationBlueprint < Blueprinter::V2::Base
  extensions << MyExtension.new
  options[:exclude_nil] = true

  field :id
end

class MyBlueprint < ApplicationBlueprint
  field :name
  object :category, CategoryBlueprint

  # inherits from "default" view
  view :extended do
    exclude :name
    collection :widgets, WidgetBlueprint[:extended]

    # inherits from "extended" view
    view :plus do
      # Uses a legacy blueprint
      object :foo, LegacyFooBlueprint, view: :plus
    end
  end
end

# inherits from "MyBlueprint extended plus" view
class MyOtherBlueprint < MyBlueprint["extended.plus"]
  ...
end

# Much like in Rails, you can define partials for your views
class WidgetBlueprint < ApplicationBlueprint
  field :name

  partial :parts do
    # Defining fields is the obvious use-case
    collection :parts, WidgetPartBlueprint
    # But they can also use the full V2 API, including extensions, views and including other partials
    extensions << MyExt.new
  end

  view :foo do
    # Include the partial
    use :parts
    field :foo
  end

  view :bar do
    # Include the partial
    use :parts
    field :bar
  end
end
```

### Rendering

```ruby
# Render the default view
MyBlueprint.render(obj, opts)
MyBlueprint[:default].render(obj, opts)

# Render a named view
MyBlueprint[:extended].render(obj, opts)
MyBlueprint["extended"].render(obj, opts)

# Render a nested view
MyBlueprint["extended.plus"].render(obj, opts)
# which is syntactic sugar for:
MyBlueprint[:extended][:plus].render(obj, opts)

# Universal render call
view = "default" # or :foo, "foo.bar", :"foo.bar", etc
MyBlueprint[view].render(obj, opts)

# It COULD be made backwards-compatible, although I dislike the "Hash soup" approach.
# Maybe have this in 1.x to ease adoption then remove in 2.0?
MyBlueprint.render(obj, view: "extended.plus")
```

### Reflection

```ruby
# Here, default refers to MyBlueprint
MyBlueprint.reflections.keys
=> [:default, :extended, :"extended.plus"]

# If you reflect from a child view, the keys are relative
MyBlueprint[:extended].reflections.keys
=> [:default, :plus]
```


jhollinger:
Alt view options:

```ruby
WidgetBlueprint.render(obj, view: "extended.plus")
WidgetBlueprint.view("extended.plus").render(obj)
WidgetBlueprint.["extended.plus"].render(obj)
```

william-stacken:
What about extractor classes that could be provided to fields and associations? Is that something that is addressed in this PR?
https://github.com/procore-oss/blueprinter/blob/b90d251d7098249b6760493a4c414db4ba920c43/lib/blueprinter/field.rb#L11

We are currently using custom extractors to cache the values of extracted fields, and to detect infinite resursion while extracting a field (field references itself). See the related issue about extractor configurability.
https://github.com/procore-oss/blueprinter/issues/404

jhollinger:
> What about extractor classes that could be provided to fields and associations? Is that something that is addressed in this PR?

We're still looking at how we want to implement extractors, and a few other things, in V2. Those will be follow-up PRs.

lib/blueprinter/v2.rb
```diff
@@ -0,0 +1,116 @@
+# frozen_string_literal: true
+
+require 'blueprinter/v2/dsl'
+require 'blueprinter/v2/options'
+require 'blueprinter/v2/reflection'
+require 'blueprinter/v2/view_builder'
+
+module Blueprinter
+  # Base class for V2 Blueprints
+  class V2
```
lessthanjacob, line 10: Reading through this again with fresh eyes, I'm less inclined to want to codify the version as part of the class itself, especially since it doesn't quite describe what the class _is_ as much. 

The versioning system we were originally going with does seem over-engineered based on our current expectations, but I'm curious if we could have a more expressive name, and then introduce those versioning semantics later on if needed (e.g. `class Blueprint` -> `Blueprint[3.0]`).

Otherwise, I'd lean more towards `BlueprintV2` if we want to stick with the explicit version in the name.

lib/blueprinter/v2.rb
```diff
@@ -0,0 +1,116 @@
...
+      attr_accessor :views, :fields, :excludes, :partials, :used_partials, :eval_mutex
+    end
+
+    self.views = ViewBuilder.new(self)
+    self.fields = {}
+    self.excludes = []
+    self.partials = {}
+    self.used_partials = []
+    self.extensions = []
+    self.options = Options.new(DEFAULT_OPTIONS)
```
lessthanjacob, line 31: **nb**: I'd expect `DEFAULT_OPTIONS` to be encapsulated solely within `Options`, and that simply calling `Options.new` would initialize it with the defaults.

lessthanjacob (commented):
I plan on giving this another thorough review, but I'm inclined to suggest that we try breaking this apart, and introduce each component separately, as there's a lot of new concepts at play here. I know we're merging into a release branch, but I believe it'd make it easier to break down iterate on moving forward.

lib/blueprinter/v2.rb
jhollinger: I've made `V2` a module and moved this code to `V2::Base` for now.

jhollinger:
For posterity (as discussed in the meeting): In my mind, the important part of this PR is the view/class inheritance structure and associated DSL. That is: how fields, associations, partials, options, and extensions are inherited from a parent class/view into child classes/views in a consistent, predictable manner, removing the need for global config.

The included implementations of `V2::Field`, `V2::Options` etc are very anemic - someone will need to take these and run with them in follow up discussions and PRs.

lessthanjacob (approved):

friscomm (approved):

Issue #438, opened 2024-07-11
Use autoload instead of require/require_relative


jhollinger:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

No

### Describe the feature you'd like to see implemented

As we progress towards dual Legacy and V2 APIs, we should use `autoload` instead of `require`/`require_relative`. Unused code can remain unloaded, and it's just easier to reason about IMHO.

Since this isn't a material change, we could technically merge this right into `main`. Might prevent some conflicts with the V2 branch as work in `main` continues. Thoughts?

jhollinger:
Done

PR #445, opened 2024-07-18
Use autoload instead of require/require_relative

jhollinger:
Checklist:

* [x] My build is green

Per #438, start using `autoload` for some things, and use `require 'blueprint/...` instead of `require_relative` in other places. This allows parts of the codebase to be loaded only if they're needed (Extensions, Transformers, V2, etc).

Based on feedback, the logic of `autoload` vs `require` is now:

- Use `autoload` for optional, top-level constants (`Base`, a future `V2`, `Extension`, `Transformer`, etc).
- Autoloaded files should `require` any constants they directly reference.
- Errors are an exception - we'll probably have an increasing number of them, so autoload makes IMHO.
- The above rules should ensure that nothing is autoloaded during render, except for errors.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -34,7 +34,7 @@ YARD::Rake::YardocTask.new do |t|
 end
 
 Rake::TestTask.new(:benchmarks) do |t|
-  t.libs << 'spec'
+  t.libs << 'spec' << 'lib'
   t.pattern = 'spec/benchmarks/**/*_test.rb'
   t.verbose = false
 end
--- a/Rakefile
+++ b/Rakefile
@@ -34,7 +34,7 @@ YARD::Rake::YardocTask.new do |t|
 end
 
 Rake::TestTask.new(:benchmarks) do |t|
-  t.libs << 'spec' << 'lib'
+  t.libs.append('lib', 'spec')
   t.pattern = 'spec/benchmarks/**/*_test.rb'
   t.verbose = false
 end
--- a/lib/blueprinter/errors.rb
+++ b/lib/blueprinter/errors.rb
@@ -2,6 +2,6 @@
 
 module Blueprinter
   module Errors
-    class InvalidBlueprint < Blueprinter::BlueprinterError; end
+    autoload :InvalidBlueprint, 'blueprinter/errors/invalid_blueprint'
   end
 end
--- a/lib/blueprinter.rb
+++ b/lib/blueprinter.rb
@@ -1,7 +1,37 @@
 # frozen_string_literal: true
 
-require_relative 'blueprinter/base'
-require_relative 'blueprinter/extension'
-
 module Blueprinter
+  # Core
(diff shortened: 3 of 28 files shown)


lib/blueprinter/errors.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
```
friscomm, line 1: The new autoload functionality is new to me, so this is likely a basic question - why is this autoload separate from the others in `lib/blueprinter.rb`?

Is it more of an organization thing? Like to avoid putting all of the required autoloads in one massive file? 

Or is there some valuable functional difference by putting it here? Do we gain anything performance-wise by separating this out?

Just trying to further my understanding as I'm learning about it :)

jhollinger: Yeah, purely organizational.

Rakefile
```diff
@@ -34,7 +34,7 @@ YARD::Rake::YardocTask.new do |t|
 end
 
 Rake::TestTask.new(:benchmarks) do |t|
-  t.libs << 'spec'
+  t.libs << 'spec' << 'lib'
```
lessthanjacob, line 37: nb:

```suggestion
  t.libs.append('lib', 'spec')
```

lessthanjacob (commented):
I'd be curious to hear your thoughts on whether we should _broadly_ use `autoload` everywhere, or if we should only focus on code/modules that will end up being optional, or truly "modular" in the future. 

I definitely agree that being able to selectively (lazily) load functionality in future updates will be a fantastic add, but I imagine we'll likely find ourselves in a place where there's still consistent, essential dependencies (e.g. `Blueprinter::Base` will always leverage `View`). In those scenarios, I'd argue that `require` captures the intent more, and should be preferred over `autoload`. 

Additionally, while our performance/efficiency margins are small within this gem, it's worth noting that we're optimizing for a smaller memory footprint and reduced start-up time at the expense of the first render. I'd argue that start-up time is less important than time to first render in the context of a production application, but I think there's a balance to be made here. 

Essentially, my **TL;DR** here is that I think we should still use `require` for core dependencies, and focus on leveraging `autoload` for optional/modular functionality (the pattern for this might not be apparent yet, but we could always revisit in the future). 

All that said, I'm open for to this approach if we're comfortable with the tradeoffs, and it aligns with our longer term vision.

jhollinger:
> I'd be curious to hear your thoughts on whether we should _broadly_ use `autoload` everywhere, or if we should only focus on code/modules that will end up being optional, or truly "modular" in the future....

@lessthanjacob I went "whole hog" b/c I didn't love the idea of loading things two different ways. You make good points, though. How about:
- `Base`
- `V2` (or however that shakes out)
- `Extension`?
- `Errors:*`?

jhollinger:
Updated with a hybrid autoload/require approach (see updated description).

lessthanjacob (approved):

PR #476, opened 2024-10-14
Extract Rendering Functionality and Remove BaseHelpers Module

### LINKED ISSUE #432: Extract Rendering Logic from `BaseHelpers`
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

No

### Describe the feature you'd like to see implemented

Currently, rendering logic is split between a `BaseHelpers` module (primarily), and `Blueprinter::Base` (with the former also being included in the latter). I believe this was done to help "slim down" `Blueprinter::Base`, but `BaseHelpers` has seemingly turned into a sort of "junk drawer" of methods. 

Since rendering is a clearly defined concept in the context of `Blueprint`, it would be helpful to encapsulate relevant logic in a specific class (e.g. `Renderer`), which would help with comprehensibility, testability, and extensibility moving forward. 

### Describe alternatives you've considered

_No response_

### Additional context

At the moment, the rendering flow looks something like so:

```mermaid
sequenceDiagram
actor User as User
participant B as Blueprinter
participant BH as BaseHelper
participant E as Blueprinter.configuration.extensions

User ->> B: .render(Object)
B ->> BH: .prepare_for_render
BH ->> BH: .validate_root_and_meta
BH ->> B: .prepare
B ->> E: .pre_render
E ->> B: Object
B ->> BH: .prepare_data
BH ->> B: Hash | Array
B ->> BH: .prepend_root_and_meta
BH ->> B: Hash | Array
B ->> B: .jsonify
B ->> User: JSON Object
```

lessthanjacob:
Resolves #432 

Checklist:

* [x] My build is green

## Changelog
- Extract rendering logic into `Rendering` module
- Remove `BaseHelpers` module
- Reformat `Base` class with `class << self` syntax
- Rename public `prepare_object` to `hashify` to better match expectations. 
- Introduce new exceptions for meta/root errors


lib/blueprinter/configuration.rb
```diff
@@ -24,18 +37,15 @@ def initialize
       @extractor_default = AutoExtractor
       @default_transformers = []
       @custom_array_like_classes = []
-    end
-
-    def extensions
-      @extensions ||= Extensions.new
+      @extensions = Extensions.new
```
jhollinger, line 40: Doesn't this mean, in the code below, that `ExtensionA` would be dropped?

```ruby
  config.extensions << ExtensionA.new
  config.extensions << ExtensionB.new
end
```

lessthanjacob: This should still work!

```ruby
irb(main):001> class ExtensionA; end
=> nil
irb(main):002> class ExtensionB; end
=> nil
irb(main):003* Blueprinter.configure do |config|
irb(main):004*   config.extensions << ExtensionA.new
irb(main):005*   config.extensions << ExtensionB.new
irb(main):006> end
=> #<Blueprinter::Extensions:0x000000014d03fa68 @extensions=[#<ExtensionA:0x000000014d03f978>, #<ExtensionB:0x000000014d03f8b0>]>
irb(main):007> Blueprinter.configuration.extensions
=> #<Blueprinter::Extensions:0x000000014d03fa68 @extensions=[#<ExtensionA:0x000000014d03f978>, #<ExtensionB:0x000000014d03f8b0>]>
irb(main):008> 
```

It's effectively the same behavior as before, just with the `@extensions` being set to `Extensions.new` _before_ the block is evaluated instead of on first `Configuration#extensions` call.

Issue #407, opened 2024-03-21
Implement association as it's own thing


sandstrom:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

Right now, association is internally implemented as a field, with a magic `:assocation` option.

### Describe the feature you'd like to see implemented

I'd implement association as a proper class internally, similar to `Field`.

```ruby
    def self.association(method, options = {}, &block)
      validate_blueprint!(options[:blueprint], method)

      make_association(
        method,
        options.merge(
          extractor: options.fetch(:extractor) { AssociationExtractor.new }
        ),
        &block
      )
    end
```

sandstrom:
Happy to discuss this in more detail, and elaborate more.

sandstrom:
Friendly ping @lessthanjacob @njbbaer @ritikesh

If you think this is a reasonable suggestion, let me know! If so, I'd propose these steps:

- I'll flesh out the proposed change in some more detail (written, in this issue)
- You get another chance to provide more feedback, and if you're happy we'll open a PR

For reference, there are also a few other issues that I've opened, where I'd also be happy to get your input.

### Full list

- https://github.com/procore-oss/blueprinter/issues/403
- https://github.com/procore-oss/blueprinter/issues/404
- https://github.com/procore-oss/blueprinter/issues/405
- https://github.com/procore-oss/blueprinter/issues/406
- https://github.com/procore-oss/blueprinter/issues/407

sandstrom:
Friendly ping @lessthanjacob @njbbaer @ritikesh

Happy to hear your thoughts on this!

ritikesh:
What would the responsibilities be of this Association class outside of how it's being handled within the ViewCollection/View contexts today?

sandstrom:
It would be a sibling class next to `Field`. It would work just the same as the `Field` class does today.

Mostly just seemed like a design that's easier to comprehend, because right now an association is a field with the "magic" option called `:association` with value `true`.

That said, I don't feel strongly about this.

An alternative would be to turn it into a property on the field (similar to what's proposed for the other magic options in https://github.com/procore-oss/blueprinter/issues/405).

ritikesh:
They do behave similarly. It's just a readability thing with no major functional differences barring associations having their own blueprinter/view based rendering.
> An alternative would be to turn it into a property on the field (similar to what's proposed for the other magic options in https://github.com/procore-oss/blueprinter/issues/405).

[...]

Issue #405, opened 2024-03-21
Separate internal options from external ones


sandstrom:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

Right now options are used both internally and externally. For example, `:block` and :association` are blessed options with internal purpose.

They are sometimes 'scrubbed out' for example this line in `AssociationExtractor`:

```ruby
options_without_default = options.except(:default, :default_if)
```

This increases the risk of colliding with user-defined options, and also make the internal code slightly harder to follow/modify.

### Describe the feature you'd like to see implemented

I would separate the two sets, in one of two ways:

1. Either into e.g. `internal_options` and `options`, and keep things such as `:block` in the internal set, or 
2. into dedicated instance variables, such as `@field_block` (see example below).

I'd prefer the second way, but (1) would at least be an improvement over the current state.

```ruby
# base.rb
def self.field(method, options = {}, &block)
  current_view << Field.new(
    method,
    options.fetch(:name) { method },
    options.fetch(:extractor) { Blueprinter.configuration.extractor_default.new },
    self,
    inline_definition: block)
  )
end

# field.rb
class Field
  attr_reader :method, :name, :extractor, :options, :blueprint, :inline_definition

  def initialize(method, name, extractor, blueprint, inline_definition: nil, options = {})
    @method = method
    @name = name
    @extractor = extractor
    @blueprint = blueprint
    @options = options
    @inline_definition = inline_definition
  end
```

sandstrom:
Happy to discuss this in more detail, and elaborate more.

lessthanjacob:
I'd definitely like to see a bit more intentionality here.`options` feels like this sort of "grab bag" of information, and I certainly agree that it increases the cognitive overhead when reading through it. 

 So in that sense, I'd also push more towards your second proposed option.

sandstrom:
Still relevant!

sandstrom:
Not stale

sandstrom:
Not stale

sandstrom:
not stale

PR #479, opened 2024-10-23
Experimental V2 renderer

jhollinger:
An **experimental** serializer/renderer for the proposed V2 base. Rapidly evolving. I know it looks large, but most of that is tests. It's broken up into small commits so we can easily pull out pieces and PR them individually if we want. Notes:

- A simple formatter DSL in Blueprints
- Extractors are extensions
- The serializer dogfoods the reflection API
- The serializer dogfoods extension hooks: all built-in options come from built-in extensions
- Several optional extensions built in (should we have a *blueprinter-contrib* repo for these?)
- Hook design solves https://github.com/procore-oss/blueprinter-activerecord/issues/30
- Can be significantly faster than V1 - up to 60%! (details below)
- Blueprints get instantiated:
  - During each render
  - A _single instance_ is shared throughout the render (e.g. a single `WidgetBlueprint` instance)
  - Field blocks, if blocks, formatter blocks, etc. are instance-evaled against the blueprint instance
  - This allows for helper methods and state during the render

### Performance

The perf characteristics of V1 and V2 vary greatly based on:

1. The total number of fields and associations in a Blueprint.
2. The relative number of fields vs associations in a Blueprint.
3. How many items are being rendered overall.

```
# 2 fields, 1 objects, 0 collections
10,000 widgets 10x: V2 50.64% faster (0.3797 sec)
1,000 widgets 100x: V2 48.85% faster (0.3582 sec)
500 widgets 100x: V2 46.48% faster (0.1675 sec)
250 widgets 100x: V2 45.56% faster (0.0819 sec)
100 widgets 250x: V2 39.26% faster (0.0707 sec)
25 widgets 500x: V2 13.27% faster (0.0121 sec)
5 widgets 1000x: V2 56.76% slower (0.0473 sec)
1 widgets 1000x: V2 87.57% slower (0.0599 sec)

# 10 fields, 5 objects, 2 collections
10,000 widgets 10x: V2 59.70% faster (4.6962 sec)
1,000 widgets 100x: V2 57.14% faster (4.1727 sec)
500 widgets 100x: V2 56.30% faster (2.0148 sec)
250 widgets 100x: V2 54.93% faster (0.9509 sec)
100 widgets 250x: V2 53.53% faster (0.9159 sec)
25 widgets 500x: V2 49.31% faster (0.4208 sec)
5 widgets 1000x: V2 25.31% faster (0.0953 sec)
1 widgets 1000x: V2 46.88% slower (0.0715 sec)

# 100 fields, 50 objects, 25 collections
10,000 widgets 10x: V2 62.32% faster (77.7100 sec)
1,000 widgets 100x: V2 69.06% faster (87.1926 sec)
500 widgets 100x: V2 68.95% faster (43.3407 sec)
250 widgets 100x: V2 68.18% faster (20.9669 sec)
100 widgets 250x: V2 60.04% faster (14.8659 sec)
25 widgets 500x: V2 62.47% faster (7.7011 sec)
5 widgets 1000x: V2 56.03% faster (2.7653 sec)
1 widgets 1000x: V2 30.53% faster (0.3391 sec)
```

Note that on the very small end of things (few items with few fields) V2 is actually slower. But we're talking fractions of a ms.


lib/blueprinter/v2.rb
sandstrom, line 1: I'm guessing this is temporary, but I wouldn't add a `V2` namespace. Instead I'd just bump the gem version.

Much less work, and less confusing, to maintain multiple versions that way.

sandstrom (commented):
Looks good!

I would consider two things that I don't see here (unless I missed it).

- Implement association as it's own thing, instead of having that be just a field. Fields are inherently different from associations, and it'll be easier to work with if that's reflected in the internal code as well. [ref](https://github.com/procore-oss/blueprinter/issues/407)
- Separate internal options from external ones, instead of having some 'blessed' or 'special' options among user-defined ones. [ref](https://github.com/procore-oss/blueprinter/issues/405)

jhollinger:
@sandstrom 

>     * Implement association as it's own thing, instead of having that be just a field. Fields are inherently different from associations, and it'll be easier to work with if that's reflected in the internal code as well. [related issue](https://github.com/procore-oss/blueprinter/issues/407)

I think this is accounted for. I've actually split "assocation" into "singular" and "multiple" variants (currently with the placeholder names `object` and `collection`). They're implemented as separate structs from field and can be fetched separately from the reflection API.

>     * Separate internal options from external ones, instead of having some 'blessed' or 'special' options among user-defined ones. [related issue](https://github.com/procore-oss/blueprinter/issues/405)

This is a little tricky given the hook-based design. To ensure that extensions can be just as powerful as most built-in functionality, all built-in options (if, unless, default, root, extractor, etc) are implemented with extension hooks. I suppose we could still store those options separately if there's a consensus around it. I _did_ move `blueprint` and `view` out of options, but for slightly different reasons.

sandstrom:
> I think this is accounted for. I've actually split "assocation" into "singular" and "multiple" variants (currently with the placeholder names object and collection). They're implemented as separate structs from field and can be fetched separately from the reflection API.

Ah, I see now. I was looking at this PR diff, but the true diff seems to be this one:

https://github.com/procore-oss/blueprinter/compare/main...jh/2.0-renderer

> This is a little tricky given the hook-based design. To ensure that extensions can be just as powerful as most built-in functionality, all built-in options (if, unless, default, root, extractor, etc) are implemented with extension hooks. But I suppose we could still store those options separately if there's a consensus around it. I did move blueprint and view out of options, but for slightly different reasons.

I'm just providing feedback here, based on some thoughts I had when we implemented Blueprinter in our codebase. 

I understand, though I guess I also slightly disagree with the idea to implement built-in functionality using extension hooks. I think I would have taken a different path, but it's also quite subjective so take this with a grain of salt 😄 

Overall, great to see the new V2 taking shape! 🎉

lessthanjacob:
Awesome work here @jhollinger! 

Since we're aligned on the general direction, could we start pulling out more targeted branches from these changes? It'd be great to tie them back to some of the long-standing issues we've had open. 

Additionally, I was hoping to jump back in and get more "hands on" again with this work, but didn't want to clobber into anything from this branch if it's accounted for here!

sandstrom:
not stale

sandstrom:
Awesome! 🎉

PR #503, opened 2025-02-16
Feature: include original view in the options for custom field block

neo87cs:
The concept is simple, let's suppose there is a need of a custom object attribute, like:
```
view :default do
  field :property do |object, options|
    { foo: 'foo'}.merge(options[:view] == :nested ? { foo2: 'foo2' } : {})
  end
end

view :nested do
  # some other fields
end
```
Currently there is no way to see which view triggered the field rendering, the PR simply add the exposition of the triggering view inside options, to avoid the need of conditional rendering based on view to duplicate the code (cause currently i need to define a duplication inside a view of the default view custom field)

Checklist:

* [x] My build is green

### DIFF
--- a/README.md
+++ b/README.md
@@ -533,7 +533,9 @@ _NOTE:_ `taskable.blueprint` should return a valid Blueprint class. Currently, `
 <summary>Defining A Field Directly In The Blueprint</summary>
 
 
-You can define a field directly in the Blueprint by passing it a block. This is especially useful if the object does not already have such an attribute or method defined, and you want to define it specifically for use with the Blueprint. This is done by passing `field` a block. The block also yields the object and any options that were passed from `render`. For example:
+You can define a field directly in the Blueprint by passing it a block. This is especially useful if the object does not already have such an attribute or method defined, and you want to define it specifically for use with the Blueprint. This is done by passing `field` a block. The block also yields the object and any options that were passed from `render` with, additionally,
+`view` which can be useful to access the view who triggered the field rendering (imagine view `a` include view `b`, during rendering of a generic field you can see view `a` on `options[:view]`).
+For example:
 
 ```ruby
 class UserBlueprint < Blueprinter::Base
--- a/README.md
+++ b/README.md
@@ -533,9 +533,7 @@ _NOTE:_ `taskable.blueprint` should return a valid Blueprint class. Currently, `
 <summary>Defining A Field Directly In The Blueprint</summary>
 
 
-You can define a field directly in the Blueprint by passing it a block. This is especially useful if the object does not already have such an attribute or method defined, and you want to define it specifically for use with the Blueprint. This is done by passing `field` a block. The block also yields the object and any options that were passed from `render` with, additionally,
-`view` which can be useful to access the view who triggered the field rendering (imagine view `a` include view `b`, during rendering of a generic field you can see view `a` on `options[:view]`).
-For example:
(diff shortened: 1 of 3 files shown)


neo87cs:
@ritikesh does it make sense? I did a rebase and as main is moving forward i would like to know if i should rebase with main and check it again

lessthanjacob:
Hey @neo87cs! Thanks for opening up this pull request (and apologies for the delayed response here)!

I'd be curious to understand your use case a bit more! I think it's reasonable to be able to introspect the `view` that is being used when rendering, but with how things work at the moment (combined with how clunky `options` handling is at the moment), there _might_ be an easier way to accomplish this currently. 

This is a fairly straightforward change, though, so if this is the best way to approach your use-case, I think it'd be reasonable to add.

README.md
```diff
@@ -533,7 +533,9 @@ _NOTE:_ `taskable.blueprint` should return a valid Blueprint class. Currently, `
 <summary>Defining A Field Directly In The Blueprint</summary>
 
 
-You can define a field directly in the Blueprint by passing it a block. This is especially useful if the object does not already have such an attribute or method defined, and you want to define it specifically for use with the Blueprint. This is done by passing `field` a block. The block also yields the object and any options that were passed from `render`. For example:
+You can define a field directly in the Blueprint by passing it a block. This is especially useful if the object does not already have such an attribute or method defined, and you want to define it specifically for use with the Blueprint. This is done by passing `field` a block. The block also yields the object and any options that were passed from `render` with, additionally,
+`view` which can be useful to access the view who triggered the field rendering (imagine view `a` include view `b`, during rendering of a generic field you can see view `a` on `options[:view]`).
```
lessthanjacob, line 537: Let's move to this a specific example below. For example:

<details>
<summary>Accessing the View Option</summary>
The `view` provided to the `render` call is implicitly available in the `options` hash within a `field` block, and can be referenced as needed.

For example:

```ruby
  identifier :uuid
  field :full_name do |user, options|
    prefix = options[:view] == :admin ? '[Admin]' : options[:title_prefix]
    "#{prefix} #{user.first_name} #{user.last_name}"
  end

  view :admin do
    field :access_level
  end
end
```

Usage:
```ruby
puts UserBlueprint.render(user, title_prefix: "Mr", view: :admin)
```
Output:
```json
{
  "full_name": "[Admin] John Doe",
  "access_level": "4"
}
```

</details>

neo87cs:
Hi @lessthanjacob ,
thank you for your reply and your effort on it, i did the commit to align to the suggestions. About our use case, we have a set of entities for which we are implement something like that:

- entity blueprinters defined a standard set of views: `small (default)`, `medium`, `large`, `xlarge`;
- starting from medium associations (they could be associations or pseudo associations for denormalized tables) are rendered coherently with correspondant view.

We couldn't achieve that with the current implementation cause we couldn't inject the correct view if we didn't write the same association rendering through each view. I though about it for a while and i think it make sense to expose this kind of facility to the custom field renderer, as it is exposed but not required to be used.

lessthanjacob (approved):

RocKhalil:
@lessthanjacob why did this PR get merged ? I have 10+ projects that can't update the gem ( failing tests ) because of this PR.

Imagine having this:
```
class UserBlueprint  < Blueprinter::Base
  field :id, :email

  view :extended do
    association :company, blueprint: CompanyBlueprint
end 
  
class CompanyBlueprint  < Blueprinter::Base
  field :id, :name

  view :extended do
    association :users, blueprint: UserBlueprint
  end
end
```

if you call `UserBlueprint.render_as_hash(current_user, view: :extended)` the `company` will be rendered as extended.
this doesn't make any sense.

Issue #480, opened 2024-10-25
[bug] Cryptic error on render when fields have both symbol and string names


branson-simplethread:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is this a regression?

- [ ] Yes, this used to work before

### Current Behavior

Title pretty much says it all. If a blueprint has both a field with a symbol name and a field with a string name, `render` raises an `ArgumentError`:

```ruby
ArgumentError: comparison of Symbol with String failed
from .../blueprinter-1.1.2/lib/blueprinter/view_collection.rb:32:in `sort_by'
```

### Expected Behavior

Blueprinter handles this case and does not raise an error, or raises a more descriptive error.

### Steps To Reproduce

Minimal test case:

```ruby
class Test < Blueprinter::Base
  field(:foo) { 1 }
  field('bar') { 2 }
end

Test.render nil
```

### Environment

```markdown
- Ruby 3.1.6
- Blueprinter 1.1.2
```

PR #505, opened 2025-02-18
Explicitly Cast Field/Association method argument to Symbol

lessthanjacob:
## Changelog
- Addresses #480. 

Checklist:

* [x] My build is green

### DIFF
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -163,12 +163,7 @@ module Blueprinter
           blueprint: options.fetch(:blueprint),
           parent_blueprint: self,
           view: options.fetch(:view, :default),
-          options: options.except(
-            :name,
-            :extractor,
-            :blueprint,
-            :view
-          ).merge(block:)
+          options: options.except(:name, :extractor, :blueprint, :view).merge(block:)
         )
       end
 
--- a/lib/blueprinter/base.rb
+++ b/lib/blueprinter/base.rb
@@ -51,7 +51,7 @@ module Blueprinter
           name,
           extractor,
           self,
-          block: block
+          block:
         )
       end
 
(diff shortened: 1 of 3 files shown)


lib/blueprinter/base.rb
```diff
@@ -51,7 +51,7 @@ def identifier(method, name: method, extractor: Blueprinter.configuration.extrac
           name,
           extractor,
           self,
-          block: block
+          block:
```
jhollinger, line 54: Is this intentional? I see it in three places, and it seems like `block` would no longer be passed.

lessthanjacob: It's a side-effect of my local `rubocop` auto-correction based on [this 3.1+ change](https://bugs.ruby-lang.org/issues/14579). 

This should function exactly as before, but let me double check that we have proper test coverage here.

lessthanjacob: I just added a spec that covers the case when an `identifier` is leveraging a block. 

How did you specify this locally? This is working for me on `3.1` (I'll wait to see if the tests pass on `3.2` as well):

```ruby
irb(main):001* class WidgetBlueprint < Blueprinter::Base
irb(main):002*   identifier :id do |object, _options|
irb(main):003*     object.id * 2
irb(main):004*   end
irb(main):005> end
=> 
#<Blueprinter::Field:0x0000000102c9ea48
 @blueprint=WidgetBlueprint,
 @extractor=
  #<Blueprinter::AutoExtractor:0x0000000102e879e0
   @block_extractor=#<Blueprinter::BlockExtractor:0x0000000102e87800>,
   @datetime_formatter=#<Blueprinter::DateTimeFormatter:0x0000000102e87760>,
   @hash_extractor=#<Blueprinter::HashExtractor:0x0000000102e87940>,
   @public_send_extractor=#<Blueprinter::PublicSendExtractor:0x0000000102e878a0>>,
 @method=:id,
 @name=:id,
 @options={:block=>#<Proc:0x0000000102e876c0 (irb):2>}>
 irb(main):006> Widget = Struct.new(:id)
=> Widget
irb(main):007> WidgetBlueprint.render(Widget.new(1))
=> "{\"id\":2}"
irb(main):008> WidgetBlueprint.render(Widget.new(2))
=> "{\"id\":4}"
irb(main):009> WidgetBlueprint.render(Widget.new(3))
=> "{\"id\":6}"
```

jhollinger: nvm, I made a dumb mistake. Carry on!

jhollinger (approved):

PR #506, opened 2025-02-18
V2 extension hooks and a hook runner

jhollinger:
Extracts V2 extension hooks from #479, and uses the new hook runner for our (only) V1 hook.

- `Blueprinter::Extension` - V2 hooks added
- `Blueprinter::Hooks` - An efficient hook runner for V1 and V2
- `Blueprinter::V2::Context:*` - The structs passed to most V2 hooks and other parts of the API

There's also a [run at documentation for the hooks](https://procore-oss.github.io/blueprinter/api/extensions.html).


lib/blueprinter/extension.rb
```diff
@@ -2,11 +2,195 @@
...
+    #
+    def blueprint_fields(_context) = []
+
+    #
+    # Modify or replace the object passed to render/render_object. The returned object is what will be rendered.
+    #
+    # @param context [Blueprinter::V2::Context]
+    # @return [Object]
+    #
+    def input_object(context) = context.object
```
lessthanjacob, line 83: What's the use case for this vs `prepare`?

lib/blueprinter/extension.rb
```diff
@@ -2,11 +2,195 @@
...
+    #
+    def input_collection(context) = context.object
+
+    #
+    # Modify or replace the object result (stored in context.value) before final render (e.g. to JSON). The returned object is what will be rendered.
+    #
+    # @param context [Blueprinter::V2::Context]
+    # @return [Object]
+    #
+    def output_object(context) = context.value
```
lessthanjacob, line 99: This might be suffering from our use of serialize/render, but could we rephrase this as `after_serialization` / `post_serialization` or something similar?

lib/blueprinter/extension.rb
```diff
@@ -2,11 +2,195 @@
...
+    #
+    def output_collection(context) = context.value
+
+    #
+    # Modify or replace an object right before it's serialized by a Blueprint. The returned object will be used as the input to the Blueprint.
+    #
+    # @param context [Blueprinter::V2::Context]
+    # @return [Object]
+    #
+    def blueprint_input(context) = context.object
```
lessthanjacob, line 115: What's a use case for this that isn't captured by a previous hook?

lib/blueprinter/extension.rb
```diff
@@ -2,11 +2,195 @@
...
+    # Instrument extension hook calls. MUST yield!
+    #
+    # @param _extension [Blueprinter::Extension] Instance of the extension
+    # @param _hook [Symbol] Name of hook being called
+    #
+    def around_hook(_extension, _hook) = yield
+
+    # If this returns true, around_hook will not be called when this extension's hooks are run. Used by core extensions.
+    def hidden? = false
+
```
lessthanjacob, line 191: Reading through all of these, it seems like we're aiming to take the stance that "hooks" drive _everything_, including our internal implementation. Is that accurate?

lib/blueprinter/hooks.rb
```diff
@@ -0,0 +1,131 @@
+# frozen_string_literal: true
+
+module Blueprinter
+  # An interface for running extension hooks efficiently
+  class Hooks
+    def initialize(extensions)
```
lessthanjacob, line 6: Can we clarify the input type here?

lib/blueprinter/hooks.rb
```diff
@@ -0,0 +1,131 @@
...
+      @hooks = Extension.
```
lessthanjacob, line 7: Let's add a private `attr_accessor` for this.

lib/blueprinter/hooks.rb
```diff
@@ -0,0 +1,131 @@
...
+        public_instance_methods(false).
```
lessthanjacob, line 8: Can we let `Extension` provide its own available hooks instead of inspecting its public methods?

lib/blueprinter/hooks.rb
```diff
@@ -0,0 +1,131 @@
...
+        end
+    end
+
+    #
+    # Checks if any hooks of the given name are registered.
+    #
+    # @param hook [Symbol] Name of hook to call
+    # @return [Boolean]
+    #
+    def has?(hook)
```
lessthanjacob, line 20:
```suggestion
    def registered?(hook)
```

lib/blueprinter/hooks.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    #
+    # Runs each hook.
+    #
+    # @param hook [Symbol] Name of hook to call
+    # @param arg [Object] Argument to hook
+    #
+    def each(hook, arg)
+      @hooks.fetch(hook).each { |ext| call(ext, hook, arg) }
+    end
```
lessthanjacob, line 32: Since we're delegating to `@hooks` for some of these methods, should we just use `Forwardable` and `Enumerable` (with some slight name changes)?

```ruby
class Hooks
  include Enumerable
  extend Forwardable
   
  def_delegators :hooks, :each, :<<

  def run(hook, arg)
    each { |extension| call(extension, hook, arg) }
  end
end

File: lib/blueprinter/v2/fields.rb [hunk 10]
```diff
@@ -27,5 +27,20 @@ module V2
...
+    #
+    # blueprint = Instance of the current Blueprint class (always)
+    # field = Field | ObjectField | Collection (optional)
+    # value = The current value of `field` or the Blueprint output (optional)
+    # object = The object currently being evaluated (e.g. passed to `render` or from an association) (optional)
+    # options = Options passed to `render` (always)
+    # instances = An InstanceCache instance for sharing instances of Blueprints and Extractors during a render (always)
+    # store = A Hash to for extensions, etc to cache render data in (always)
+    #
+    Context = Struct.new(:blueprint, :field, :value, :object, :options, :instances, :store)
```
lessthanjacob, line 44: I was playing around with this concept a bit separately, but it feels like we should pull this out from the `Fields` class and iterate on it. It still feels like a bit of a "grab bag", so we could try and disambiguate use cases (e.g. `field`-level context vs `object`-level context vs. `render`-level context).

spec/extensions/extension_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+describe Blueprinter::Extension do
+  subject { Class.new(described_class) }
+
+  context 'hooks' do
+    let(:blueprint) { Class.new(Blueprinter::V2::Base) }
+    let(:field) { Blueprinter::V2::Field.new(name: :foo, from: :foo) }
+    let(:object) { { foo: 'Foo' } }
+    let(:context) { Blueprinter::V2::Context }
+
+    it 'should default collections? to false' do
```
lessthanjacob, line 12: Super *nit*, but I'd suggest omitting the "shoulds" in `it` assertions (e.g. use `it 'defaults collections? to false'`).

lib/blueprinter/extension.rb
```diff
@@ -2,11 +2,195 @@
...
+  #      - blueprint_input
+  #      - field_value
+  #      - exclude_field?
+  #      - object_value
+  #      - exclude_object?
+  #      - collection_value
+  #      - exclude_collection?
+  #      - blueprint_output
+  #    - output_object | output_collection
+  #    - json
```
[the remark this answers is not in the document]
jhollinger: It's a side-effect of the implementation.

lib/blueprinter/extension.rb
```diff
@@ -2,11 +2,195 @@
...
   #
   class Extension
     #
-    # Called eary during "render", this method receives the object to be rendered and
+    # Returns true if the given object should be treated as a collection (i.e. supports `map { |obj| ... }`).
+    #
+    # @param _object [Object]
+    # @return [Boolean]
+    #
+    def collection?(_object) = false
```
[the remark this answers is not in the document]
jhollinger: I don't think we'll be able to programmatically determine it outside of checking for Array|Set|Enumerator. `blueprinter-activerecord` could use this to add `ActiveRecord::Relation`, which I don't think we can reasonably detect another way. Happy to be wrong, though.

lib/blueprinter/hooks.rb
jhollinger: Could be reading this wrong, but I don't think it does quite the same thing. We're only delegating to a key inside `@hooks`, not `@hooks` itself.

lib/blueprinter/extension.rb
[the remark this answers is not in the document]
lessthanjacob: When we initially were looking into this, I was leaning towards something like `object.is_a?(Enumerable) && !object.is_a?(Hash)`. Outside of adding a restriction that a `Hash` can't be treated as a "collection", that seems like it'd account for the majority use cases, and I think it'd be worth enforcing that an object conforms to the `Enumerable` interface over allowing this to be totally customizable. 

I think _more_ customization is good, but we should be mindful just how far we take it IMO (to mitigate friction with adoption and the surface area for potential bugs).

lib/blueprinter/hooks.rb
jhollinger: Switched to a constant with an array of hooks.

lib/blueprinter/extension.rb
jhollinger: More or less. Partly it was an attempt to ensure we had _useful_ hooks, and partly an attempt to cover many of the feature issues/PRs we rejected as too niche.

lib/blueprinter/v2/fields.rb
jhollinger: So I was able to break it apart into five structs under a `Context` module: `Render | Object | Field | Value | Result` (`context.rb` has some docs/comments). I believe it covers all the use-cases from the experimental serializer. Thoughts?

jhollinger:
I removed the field_value, object_value, and collection_value hooks since Extractors cover those cases. We can add them back if there's truly a need.

PR #511, opened 2025-03-13
Inherit views (V2)

jhollinger:
### Changes:

- V2 blueprints inherit views from parents (like Legacy/V1)
- `view` options to skip inheriting certain attributes from the parent (i.e. fields, options, extensions)
- Freeze a blueprint/view's extensions, options and fields after it's first eval'd
- Disallow defining a `:default` view
- If a view is redefined, append to the existing definition instead of replacing it (like Legacy/V1)
  - If they **want** to replace it, they can do so using `view`'s new field/options/extension options

### Legacy/V1's "identifier" view can be emulated like this:

```ruby
class ApplicationBlueprint < Blueprinter::V2::Base
  field :id

  # without `fields: false`, this view would inherit fields from the top-level of child Blueprints
  view :identifier, fields: false do
    field :id
  end
end
```

### DIFF
--- a/lib/blueprinter/v2/base.rb
+++ b/lib/blueprinter/v2/base.rb
@@ -33,7 +33,7 @@ module Blueprinter
 
       # Initialize subclass
       def self.inherited(subclass)
-        subclass.views = ViewBuilder.new(subclass)
+        subclass.views = views.dup_for(subclass)
         subclass.schema = schema.transform_values(&:dup)
         subclass.excludes = []
         subclass.partials = partials.dup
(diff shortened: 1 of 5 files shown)


siba2893:
This looks amazing! Waiting for this to get merged

jhollinger:
Take another at the `fields` options (maybe `inherit_fields`). Use existing setters to clear options & extensions.

lessthanjacob (approved):

ritikesh:
@jhollinger / @lessthanjacob -
Should we instead just error it out during init? IMO, they're maintenance problems when people are able to re-define/patch views.
> * If a view is redefined, append to the existing definition instead of replacing it (like Legacy/V1)

PS: Sorry for coming late to this one! 😅

PR #546, opened 2025-11-20
Fix for Symbol#to_proc ArgumentError issue

tob1k:
Fixes the long-standing issue where using Symbol#to_proc syntax (e.g., `field :name, &:first_name`) would raise an ArgumentError.

## Problem:
BlockExtractor was calling all blocks with 2 arguments: `block.call(object, local_options)`. Symbol#to_proc forwards ALL arguments to the target method, causing it to call `object.method_name(local_options)`, which fails when methods don't accept extra arguments.

## Solution:
Detect Symbol#to_proc by its characteristic parameter signature `[[:req], [:rest]]` and call it with only the object argument, while regular blocks continue to receive both arguments.

### Benefits:
- Enables cleaner, more idiomatic Ruby syntax
- Eliminates need for rubocop disable comments
- Maintains full backward compatibility with existing code
- No breaking changes to API

Related: https://github.com/rubocop/rubocop/issues/8201

Checklist:

* [x] My build is green


lessthanjacob (approved):
Thanks for the contribution!

lib/blueprinter/extractors/block_extractor.rb
```diff
@@ -6,7 +6,31 @@ module Blueprinter
...
+    end
+
+    private
+
+    def symbol_to_proc?(block)
+      # Symbol#to_proc has a characteristic signature:
+      # - Parameters: [[:req], [:rest]] (one required + rest args)
+      # - This is different from regular blocks which typically have
+      #   optional parameters like [[:opt, :obj], [:opt, :options]]
+      block.parameters == [[:req], [:rest]]
```
incubus, line 33: `[[:req], [:rest]]` allocates a new array on every call. Worth extracting to a constant to reduce unnecessary allocations.

Issue #554, opened 2025-12-23
[bug] Incorrect view used during render


jhollinger:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is this a regression?

- [x] Yes, this used to work before

### Current Behavior

I've traced this issue back to #503. One could debate whether it's a "bug" or not, but it's hitting our own codebase really hard.

```ruby
class WidgetBlueprint < Blueprinter::Base
  fields :id, :name

  view :extended do
    field :code do |widget, options|
      code = options.dig(:code_cache, widget.code_id) || NullCode
      # CodeBlueprint better have whatever view is set in options!
      CodeBlueprint.render_as_hash(code, **options)
    end
  end
end

class CodeBlueprint < Blueprinter::Base
  fields :id, :name
end

# Raises 'Blueprinter::BlueprinterError: View 'extended' is not defined' on CodeBlueprint
WidgetBlueprint.render(widget, view: :extended)
```

IMHO the above is an anti-pattern, but realistically it's pretty common. What if we put the view name in a different, unused option key (e.g. `:view_name`, `:parent_view`)?

### Steps To Reproduce

See above

### Environment

```markdown
Any
```

PR #559, opened 2026-01-06
Fix View Passthrough + Extractor Cleanup

lessthanjacob:
## Changelog
* Passes the current view as `current_view` instead of `view` when extracting data for a `field`. This prevents the view from being re-used if `options` are passed along to another blueprint within a block (solves #554). 
* Adds missing specs for extractors. 

Checklist:

* [ ] I have signed off all my commits as required by [DCO](https://github.com/procore-oss/blueprinter/blob/main/CONTRIBUTING.md)
* [ ] My build is green


jhollinger (approved):

lessthanjacob:
Closing in favor of retaining the current behavior.

PR #590, opened 2026-05-21
feat(identifier): add :as option for renaming keys

OlegChuev:
Checklist:

* [x] My build is green

### DIFF
--- a/spec/integrations/shared/base_render_examples.rb
+++ b/spec/integrations/shared/base_render_examples.rb
@@ -20,6 +20,16 @@ shared_examples 'Base::render' do
       end
       it('returns json with an identifier') { should eq('{"id":"THE ID IS ' + obj_id + '"}') }
     end
+
+    context 'with :as option to rename identifier' do
+      let(:result) { '{"uuid":' + obj_id + '}' }
+      let(:blueprint) do
+        Class.new(Blueprinter::Base) do
+          identifier :id, as: :uuid
+        end
+      end
+      it('returns json with a renamed identifier') { should eq(result) }
+    end
   end
 
   context 'Given blueprint has ::field' do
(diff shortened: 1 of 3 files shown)


spec/integrations/shared/base_render_examples.rb
```diff
@@ -71,6 +81,28 @@
...
+  context 'Given blueprint has ::field with both :name and :as arguments' do
+    let(:result) { '{"first_name":"Meg","identifier":' + obj_id + '}' }
+    let(:blueprint) do
+      Class.new(Blueprinter::Base) do
+        field :id, name: :identifier, as: :ignored
+        field :first_name
+      end
+    end
+    it(':name takes precedence over :as') { should eq(result) }
+  end
```
kkohrt, line 104: To be clear, this is providing an additional argument that has the [same functionality](https://github.com/procore-oss/blueprinter/pull/590/changes#diff-72c4ffac5fd2bdecf764e5fa5272428b4c12afab57829fe21e9764fed6603f37R73-R82) as the `:name` argument, but uses the keyword `:as` instead of `:name`, right?

kkohrt (commented):
I appreciate the backwards compatibility aspect of this code change, but I am a little concerned about having two arguments that provide the same functionality, with one take precedence over another. That feels like it leaves room for confusion and/or accidental bugs being introduced into a system. If there is strong community support then it might make sense to allow a more general options hash for renaming, or else a configuration option to make it explicit what attribute name should be used.  So I am curious as to the motivation for this change, as `:name` is a [pretty common](https://github.com/search?q=repo%3Aprocore-oss%2Fblueprinter%20%3Aname&type=code) attribute name, but `:as` is [not use anywhere](https://github.com/search?q=repo%3Aprocore-oss%2Fblueprinter+%2F%3Aas%5B%5Ea-z%5D%2F&type=code).

